PR #1, opened 2023-09-20
feat: better private properties handling

nikophil:


tests/AutoMapperBaseTest.php
```diff
@@ -33,7 +33,7 @@ protected function setUp(): void
         $this->buildAutoMapper();
     }
 
-    protected function buildAutoMapper(bool $allowReadOnlyTargetToPopulate = false): AutoMapper
+    protected function buildAutoMapper(bool $allowReadOnlyTargetToPopulate = false, bool $mapPrivatePropertiesAndMethod = true, string $classPrefix = 'Mapper_'): AutoMapper
```
Korbeil, line 36: Shouldn't we put `$mapPrivatePropertiesAndMethod` at `false` by default ? (Like it is in `AutoMapper` class static method)

Korbeil (approved):

tests/AutoMapperBaseTest.php
nikophil: I think you're right, I leaved this to `true` because it was easier 😅
But I don't even think that much tests will fail if I change this

nikophil:
```ERRORS!
Tests: 167, Assertions: 455, Errors: 6, Failures: 10.
```

Korbeil:
Thanks for your contribution @nikophil ! :pray:

PR #6, opened 2023-09-27
chore(doc): document all ast code by explaining what it generates

joelwurtz:
This is mainly to help contributors to have a better understanding at what the AST generate into our different classes


Korbeil (approved):
Thanks for all the details about AST :pray:

PR #7, opened 2023-10-03
fix: `NullableTransformer` should check if array key exists

nikophil:
when mapping from array to an object, if a property is nullable and is not present in the original object, there should be no error if the property defaults to null. 

currently, this code:
```php
readonly class SomeClass
{
    public function __construct(public int|null $foo = null)
    {
    }
}

$autoMapper->map([], ClassWithNullablePropertyInConstructor::class);
```

generates the following code:
```php
// ...

            } else {
                $value_1 = null;
                if (isset($value['foo'])) {
                    $value_1 = $value['foo'];
                }
                $constructarg = $value_1;
            }

// ...
```

and results in an error where `foo` key does not exist, although we provided a default error.

At first, I wanted to add an `isset()` only if the property has a default value, but it appears that it is a little bit complex: `NullableTransformerFactory` would need to access to property info to get the write info of the property... I think this would be overkill for the case where the property is nullable and adding a `isset()` everywhere we have a array access is enough and not harmful.


Korbeil (approved):
I fear it can break some user behavior but since there was no issues with current tests, let's merge this :+1: 

Thanks for the contribution @nikophil !

nikophil:
> I fear it can break some user behavior but since there was no issues with current tests, let's merge this 👍

I don't think so:
- property is not nullable: this PR is not involved
- property is nullable, and key is provided: no problem as it used to be
- property is nullable but input is not an array: this PR is not involved

If property key is missing in the input array:
- property is nullable and has default null: this is the specific use case I wanted to fix, the property will be set to null
- property is nullable and has default to something else than null: the property will be set to null (see #8)
- property is nullable and has no default: in this case, it used to create an error, but now the property is defaulted to null. But I don't think how this could be a problem

PR #9, opened 2023-10-13
fix: mapPrivatePropertiesAndMethod should not be mandatory

nikophil:
`mapPrivatePropertiesAndMethod` machanism was buggy, this is now fixed

### DIFF
--- a/src/Extractor/FromSourceMappingExtractor.php
+++ b/src/Extractor/FromSourceMappingExtractor.php
@@ -87,7 +87,7 @@ final class FromSourceMappingExtractor extends MappingExtractor
                 $this->getMaxDepth($mapperMetadata->getSource(), $property),
                 $this->isIgnoredProperty($mapperMetadata->getSource(), $property),
                 $this->isIgnoredProperty($mapperMetadata->getTarget(), $property),
-                PropertyReadInfo::VISIBILITY_PUBLIC === ($this->readInfoExtractor->getReadInfo($mapperMetadata->getSource(), $property)?->getVisibility() ?? true)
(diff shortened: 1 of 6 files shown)


src/Extractor/FromSourceMappingExtractor.php
```diff
@@ -87,7 +87,7 @@ public function getPropertiesMapping(MapperMetadataInterface $mapperMetadata): a
                 $this->getMaxDepth($mapperMetadata->getSource(), $property),
                 $this->isIgnoredProperty($mapperMetadata->getSource(), $property),
                 $this->isIgnoredProperty($mapperMetadata->getTarget(), $property),
-                PropertyReadInfo::VISIBILITY_PUBLIC === $this->readInfoExtractor->getReadInfo($mapperMetadata->getSource(), $property)?->getVisibility() ?? true,
+                PropertyReadInfo::VISIBILITY_PUBLIC === ($this->readInfoExtractor->getReadInfo($mapperMetadata->getSource(), $property)?->getVisibility() ?? PropertyReadInfo::VISIBILITY_PUBLIC),
```
nikophil, line 90: :hand_over_mouth: 

this would have been detected by phpstan :innocent:

src/MapperMetadata.php
```diff
@@ -229,7 +229,8 @@ private function buildPropertyMapping(): void
                 null,
                 new CallbackTransformer($property),
                 $property,
-                false
+                false,
+                isPublic: true,
```
nikophil, line 233: because we're in presence of custom mapping, I think we should consider it as public, other wise `PropertyMapping::shouldIgnoreProperty()` will return true

Korbeil (approved):
LGTM, a quick note in CHANGELOG ?

PR #10, opened 2023-10-13
feat: introduce custom transformers

nikophil:


src/Transformer/CustomTransformer/CustomTransformersRegistry.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @param Type[] $sourceTypes
+     * @param Type[] $targetTypes
+     *
+     * @return null|class-string<CustomTransformerInterface>
+     */
+    public function getCustomTransformerClass(MapperMetadataInterface $mapperMetadata, array $sourceTypes, array $targetTypes, string $propertyName): string|null
+    {
+        // todo: what to do with union types?
```
Korbeil, line 31: I don't understand exactly the issue here, could you tell me more about it ?

src/Extractor/CustomTransformerExtractor.php
```diff
@@ -0,0 +1,71 @@
...
+    public function extract(string $customTransformerClass, Expr $sourceExpression): Expr
+    {
+        $statements = $this->parser->parse(file_get_contents($filename = (new \ReflectionClass($customTransformerClass))->getFileName()));
+
+        $namespaceStatement = self::findUnique(Stmt\Namespace_::class, $statements, $filename);
+        /** @var Stmt\Class_ $classStatement */
+        $classStatement = self::findUnique(Stmt\Class_::class, $namespaceStatement->stmts, $filename);
+
+        $transformMethod = $classStatement->getMethod('transform') ?? throw new \LogicException(
+            "Cannot find method \"transform()\" in custom transformer class \"{$customTransformerClass}\"."
```
Korbeil, line 39: Shouldn't exists since the interface makes that method mandatory. Is it really required ?

src/Generator/Generator.php
```diff
@@ -234,14 +238,27 @@ public function generate(MapperGeneratorMetadataInterface $mapperGeneratorMetada
...
+//                $output = new Expr\MethodCall(
+//                    new Expr\Variable('customTransformerRegistry'),
+//                    'transform',
+//                    [
+//                        new Arg(new Scalar\String_($customTransformer)),
+//                        new Arg($fieldValueVariable),
+//                    ]
+//                );
+                $propStatements = [];
+                $output = $this->customTransformerExporter->extract($transformer, $fieldValueVariable);
```
Korbeil, line 257: A quick comment to explain this is when you have a CustomTransformer and needs to parse the transformer extract method to know what to add within the Transformer

src/Generator/Generator.php
```diff
@@ -634,8 +655,16 @@ private function getCreateObjectStatements(MapperGeneratorMetadataInterface $map
...
                 $constructVar = new Expr\Variable($uniqueVariableScope->getUniqueName('constructArg'));
 
-                /* Get extract and transform statements for this property */
-                [$output, $propStatements] = $propertyMapping->transformer->transform($propertyMapping->readAccessor->getExpression($sourceInput), $constructVar, $propertyMapping, $uniqueVariableScope);
+                $fieldValueExpr = $propertyMapping->readAccessor->getExpression($sourceInput);
+
+                if (is_string($propertyMapping->transformer)) {
+                    $propStatements = [];
+                    $output = $this->customTransformerExporter->extract($propertyMapping->transformer, $fieldValueExpr);
+                } else {
```
Korbeil, line 663: Quick comment here too

src/Transformer/CustomTransformer/CustomTransformersRegistry.php
nikophil: I want the interface to stay simple. ie: I don't want it to be `public function supports(array<Type> $sources, array<Type> $targets): bool;` and to need to loop oever types

so, what to do if we map a union type to another union type, and several transformers supports the transformation?

src/Transformer/CustomTransformer/CustomPropertyTransformerInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AutoMapper\Transformer\CustomTransformer;
+
+use Symfony\Component\PropertyInfo\Type;
+
+interface CustomPropertyTransformerInterface extends CustomTransformerInterface
```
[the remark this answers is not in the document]
nikophil: > I would use something like $sourcePropertyName to not confuse things when target property name is different because of some NameConverter

This makes me raise two questions:
- shouldn't we somehow introduce the same mechanism with `targetPropertyName`?
-  currently we're actually passing the "source property" to the `transform()` method, I'm wondering if we shouldn't pass the whole `$source` object in order to be able to access to whole source object in the tranformation (ie: we could need `$source->year`, `$source->month`, `$source->day` in order to generate `$target->birthDate`

WDYT?

src/AutoMapper.php
```diff
@@ -151,6 +155,11 @@ public function bindTransformerFactory(TransformerFactoryInterface $transformerF
         }
     }
 
+    public function bindCustomTransformer(CustomTransformerInterface $customTransformer): void
```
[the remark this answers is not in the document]
nikophil: why don't we add it to `AutoMapperInterface` ? this is the one we inject in symfony's projects

BTW, adding a method to a public interface is a BC break...

src/Extractor/CustomTransformerExtractor.php
nikophil: this makes type system happy since `ClassLike::getMethod()` returns a nullable, and this is a way to fail fast if anything goes bad... when dealing with AST code, I'd rather be defensive...

src/AutoMapper.php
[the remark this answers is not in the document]
nikophil: it has been added to `MapperGeneratorMetadataRegistryInterface`

src/Extractor/AstExtractor.php
```diff
@@ -63,16 +63,6 @@ public function extract(string $class, string $method, array $inputParameters):
             throw new InvalidArgumentException("Input parameters and method parameters in class \"{$class}\" do not match.");
         }
 
-        foreach ($classMethod->getParams() as $key => $parameter) {
```
nikophil, line 66: ce check n'est pas utile, cf les tests dans `AstExtractorTest` que j'ai ajoutés

src/Extractor/AstExtractor.php
nikophil, line 1: I think we should rename this class into something like `ClassMethodToCallbackExtractor` or something like this... WDYT?

src/Transformer/CustomTransformer/CustomTransformerInterface.php
```diff
@@ -0,0 +1,20 @@
...
+interface CustomTransformerInterface
+{
+    /**
+     * @param mixed $input the source input (object or array) on which the custom transformation applies
+     *
+     * Return value depends on the custom transformer's type:
+     * For CustomModelTransformerInterface, you should return a full target object,
+     * for CustomPropertyTransformerInterface, only the target property should be returned
+     */
+    public function transform(mixed $input): mixed;
```
Korbeil, line 19: Since it's always the `source` data that is put, why not naming it `source` ?

src/Transformer/CustomTransformer/CustomPropertyTransformerInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @experimental Interface contract will change once https://github.com/symfony/symfony/pull/52510 will be merged.
+ */
+interface CustomPropertyTransformerInterface extends CustomTransformerInterface
+{
+    /**
+     * @param 'array'|class-string $source
+     * @param 'array'|class-string $target
+     */
+    public function supports(string $source, string $target, string $propertyName): bool;
```
Korbeil, line 36: Shouldn't it be an array for both `source` & `target` parameters ?

src/Transformer/CustomTransformer/CustomTransformerInterface.php
nikophil: haha I've done that is the following PR

src/Transformer/CustomTransformer/CustomPropertyTransformerInterface.php
nikophil: since we're at "property level", we're using types from `MapperMetadata`'s source and target.
See https://github.com/jolicode/automapper/pull/10/files#diff-f42c4b3c8fc2137483ba833fa3b79ab8bd81932296df954c775997278f22a4bdR39

I don't think we can have an array here. Or... I've hooked at the wrong place for custom property transformer

Korbeil:
Thanks for your contribution @nikophil ! This is a great start to improve DX within AutoMapper ! <3

Korbeil (approved):

PR #11, opened 2023-10-16
Add phpstan and resolve for level 5

Korbeil:
This PR adds phpstan with a level 5 (for now)

### DIFF
--- a/src/MapperInterface.php
+++ b/src/MapperInterface.php
@@ -19,5 +19,5 @@ interface MapperInterface
      *
      * @return mixed The mapped value
      */
-    public function &map($value, array $context = []);
+    public function &map(mixed $value, array $context = []): mixed;
 }
--- a/src/Transformer/ChainTransformerFactory.php
+++ b/src/Transformer/ChainTransformerFactory.php
@@ -11,7 +11,7 @@ use AutoMapper\MapperMetadataInterface;
  */
 final class ChainTransformerFactory implements TransformerFactoryInterface
 {
-    /** @var TransformerFactoryInterface[] */
+    /** @var array<int, list<TransformerFactoryInterface>> */
     private array $factories = [];
 
     /** @var TransformerFactoryInterface[]|null */
@@ -39,7 +39,7 @@ final class ChainTransformerFactory implements TransformerFactoryInterface
     {
         $this->sortFactories();
 
-        $transformerFactoryClass = \get_class($transformerFactory);
+        $transformerFactoryClass = $transformerFactory::class;
         foreach ($this->sorted ?? [] as $factory) {
             if (is_a($factory, $transformerFactoryClass)) {
                 return true;
--- a/src/Transformer/DateTimeTransformerFactory.php
+++ b/src/Transformer/DateTimeTransformerFactory.php
@@ -79,7 +79,7 @@ final class DateTimeTransformerFactory extends AbstractUniqueTypeTransformerFact
         return true;
     }
 
-    private function getClassName(Type $type): ?string
+    private function getClassName(Type $type): string
     {
         if (\DateTimeInterface::class !== $type->getClassName()) {
             return \DateTimeImmutable::class;
--- a/src/Transformer/SymfonyUidTransformerFactory.php
+++ b/src/Transformer/SymfonyUidTransformerFactory.php
@@ -14,6 +14,7 @@ use Symfony\Component\Uid\Ulid;
  */
 final class SymfonyUidTransformerFactory extends AbstractUniqueTypeTransformerFactory implements PrioritizedTransformerFactoryInterface
 {
+    /** @var array<string, array{0: bool, 1: bool}> */
     private array $reflectionCache = [];
 
     protected function createTransformer(Type $sourceType, Type $targetType, MapperMetadataInterface $mapperMetadata): ?TransformerInterface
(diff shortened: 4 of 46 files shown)


src/MapperInterface.php
```diff
@@ -19,5 +19,5 @@ interface MapperInterface
      *
      * @return mixed The mapped value
      */
-    public function &map($value, array $context = []);
+    public function &map(mixed $value, array $context = []): mixed;
```
nikophil, line 22: isn't this dangerous?
since this interface is implemented in user land in generated mappers.
If hot reload is disabled couldn't some problem occur?

https://3v4l.org/JtHhF

src/Transformer/ChainTransformerFactory.php
```diff
@@ -11,7 +11,7 @@
  */
 final class ChainTransformerFactory implements TransformerFactoryInterface
 {
-    /** @var TransformerFactoryInterface[] */
+    /** @var array<TransformerFactoryInterface[]> */
```
nikophil, line 14:
```suggestion
    /** @var array<int, list<TransformerFactoryInterface>> */
```
:innocent:

src/Transformer/DateTimeTransformerFactory.php
```diff
@@ -59,8 +59,8 @@ protected function createTransformerForSource(Type $targetType, MapperMetadataIn
 
     protected function createTransformerForTarget(Type $sourceType, Type $targetType, MapperMetadataInterface $mapperMetadata): ?TransformerInterface
     {
-        if (Type::BUILTIN_TYPE_STRING === $sourceType->getBuiltinType()) {
-            return new StringToDateTimeTransformer($this->getClassName($targetType), $mapperMetadata->getDateTimeFormat());
+        if (Type::BUILTIN_TYPE_STRING === $sourceType->getBuiltinType() && null !== ($targetClassName = $this->getClassName($targetType))) {
```
nikophil, line 62: not sure to understand this change, if `geClassName()` cannot return null

src/Transformer/DateTimeTransformerFactory.php
```diff
@@ -79,7 +79,7 @@ private function isDateTimeType(Type $type): bool
         return true;
     }
 
-    private function getClassName(Type $type): ?string
+    private function getClassName(Type $type): string
```
nikophil, line 82: ...buuuuut I don't understand this change since `Type::getClassName()` can return null :sweat_smile: 

if we kinda now that `$sourceType` represents an object, thanks to `isDateTimeType()`, maybe we can inform PHPStan?
```php
/**
 * @phpstan-assert-if-true !null $type->getClassName()
 */
private function isDateTimeType(Type $type): bool
```

src/Transformer/EnumTransformerFactory.php
```diff
@@ -38,12 +38,14 @@ private function isEnumType(Type $type, bool $backed = false): bool
             return false;
         }
 
-        if (!is_subclass_of($type->getClassName(), \UnitEnum::class)) {
-            return false;
-        }
+        if (null !== $type->getClassName()) {
```
nikophil, line 41: this change is strange... `is_subclass_of()` accepts `mixed`, thus `null`

https://3v4l.org/Av3ia

src/Transformer/EnumTransformerFactory.php
```diff
@@ -20,7 +20,7 @@ protected function createTransformer(Type $sourceType, Type $targetType, MapperM
         }
 
         // target is enum, source isn't
-        if (!$this->isEnumType($sourceType) && $this->isEnumType($targetType, true)) {
+        if (!$this->isEnumType($sourceType) && $this->isEnumType($targetType, true) && null !== $targetType->getClassName()) {
```
nikophil, line 23:
```suggestion
        if (!$this->isEnumType($sourceType) && $this->isEnumType($targetType, true)) {
```

and let's add
```php
/**
 * @phpstan-assert-if-true !null $type->getClassName()
 */
```
to `isEnumType()`

src/Transformer/ObjectTransformerFactory.php
```diff
@@ -53,7 +53,7 @@ private function isObjectType(Type $type): bool
             return false;
         }
 
-        if (is_subclass_of($type->getClassName(), \UnitEnum::class)) {
+        if (null !== $type->getClassName() && is_subclass_of($type->getClassName(), \UnitEnum::class)) {
```
nikophil, line 56: same as above, that's strange phpstan reports this as an error :thinking:

src/Transformer/SymfonyUidTransformerFactory.php
```diff
@@ -29,7 +30,7 @@ protected function createTransformer(Type $sourceType, Type $targetType, MapperM
             return new SymfonyUidToStringTransformer($this->reflectionCache[$sourceType->getClassName()][1]);
         }
 
-        if ($isTargetUid) {
+        if ($isTargetUid && null !== $targetType->getClassName()) {
```
nikophil, line 33: `@phpstan-assert-if-true !null $type->getClassName()` on `isUid()` :stuck_out_tongue_closed_eyes:

nikophil (changes requested):
nice improvement :+1: 

from this point I would have changed to level max and generated a baseline :innocent:

src/MapperInterface.php
Korbeil: I always expect users to reload their transformers from one version to another that's why I think this shouldn't be an issue.

src/Transformer/DateTimeTransformerFactory.php
Korbeil: I didn't know about this trick with phpstan ! It seems really nice, thanks for sharing :pray

src/MapperInterface.php
nikophil: Couldn't hot reload strategy be based on the library version?
I've never checked this part of code since I've based my code on the cache warmer

nikophil: Couldn't hot reload strategy be based on the library version?
I've never checked this part of code since I've based my code on the cache warmer

nikophil (approved):

src/MapperInterface.php
Korbeil: Today the hot reload is based onto a hash of the transformer metadata. We totally could add the AutoMapper version into this to make it force reload if the version doesn't match :100:

PR #21, opened 2023-12-09
Add VERSION constants within AutoMapper class and use it to cache transformers

### LINKED ISSUE #13: Add AutoMapper version within Mappers hash to force hot-reload
Today we have mappers that are generated based on their metadata. For each mapper we make a hash that is used to check if we need to re-generate it or not. This behavior is called hot-reload. It's enabled by default because in dev you want that feature.

Depending on the AutoMapper version, you probably have some changes on the mappers and want them to be re-generated that's why we think it would be nice to have the AutoMapper version within the generated hash so when the library version changes, it will force a mapper generation (see https://github.com/jolicode/automapper/pull/11#discussion_r1367761251).

Korbeil:
Fixes #13

### DIFF
--- a/.github/scripts/check-version.sh
+++ b/.github/scripts/check-version.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+LAST_TAG=`git describe --tags --abbrev=0`
+IFS=. components=(${LAST_TAG##*-})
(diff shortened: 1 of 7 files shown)


ternel (approved):

.github/scripts/check-version.sh
nikophil, line 1: 👍

nikophil: actually... does it work? we're currently on 8.0.2 and `VERSION = '8.1.0-DEV'` :thinking:

Korbeil: it does, otherwise CI would be red

nikophil (approved):

nikophil (approved):

PR #25, opened 2023-12-18
feat: pass full input object to property custom transformers

nikophil:


Korbeil (approved):
Could you add something into CHANGELOG about the changes and some more documentation on how to use this out of the box (meaning without Symfony integration for now) ?
Otherwise it seems all fine to me :ok_hand:

docs/homepage.md
```diff
@@ -76,4 +76,71 @@ dump($automapper->map(new InputUser('John', 'Doe', 28), DatabaseUser::class));
...
+
+```php
+use Symfony\Component\PropertyInfo\Type;
+
+class InputUserToDatabaseUserCustomTransformer implements CustomModelTransformerInterface
+{
+    public function supports(array $sourceTypes, array $targetTypes): bool
+    {
+        return $this->hasType($sourceTypes, DatabaseUser::class) && $this->hasType($targetTypes, OutputUser::class);
+    }
```
nikophil, line 121: I must maintain that I really dislike this and I think it is a terrible DX. 

From a user perspective, If I want to create a custom mapper between `DatabaseUser` and `OutputUser`, I expect the method `support()` to be passed two string, and not two arrays of strings :shrug:

Korbeil:
Thanks for the contribution @nikophil :pray:

PR #26, opened 2023-12-20
Fix mappings involving DateTimeInterface type

priyadi:
A mapping from `DateTimeInterface $timeCreated` to `DateTimeInterface $timeCreated` will result in an error if the source variable is an instance of `DateTimeImmutable`:

DateTimeImmutable::createFromMutable(): Argument #1 ($object) must be of type DateTime, DateTimeImmutable given

This fixes the problem by using `DateTimeImmutable::createFromInterface`.

Also included complete permutations between all the possible `DateTime` types.


Korbeil:
Thanks a lot for your first PR @priyadi :tada: 
Could you fix CS ? You can run it locally by doing: `tools/php-cs-fixer/vendor/bin/php-cs-fixer fix --config=tools/php-cs-fixer/.php-cs-fixer.php` (You have to `composer update` inside `tools/php-cs-fixer/` first)

priyadi:
@Korbeil done

Korbeil:
> @Korbeil done

Could you also add a note in the CHANGELOG about your changes ? :pray:

Korbeil (approved):
Thanks a lot for your contribution ! :heart:

PR #31, opened 2024-01-10
Compability with PhpParser v5

Korbeil:
Allow compatibility with `nikic/php-parser` v5
This is just a compatibility layer there is no BC


Korbeil:
Closed in favor of #36

PR #33, opened 2024-01-12
fix: allow usage of imported class names in custom transformers

### LINKED ISSUE #32: [CustomTransformer] Find & resolve variables TypeContext within an extracted `transform` method
Hello,
Let's take this example:
```php
<?php

use App\AddressDTO;

final readonly class FooTransformer implements CustomModelTransformerInterface
{
    public function supports(array $sourceTypes, array $targetTypes): bool
    {
        return true;
    }

    public function transform(mixed $input): mixed
    {
        $addressDTO = new AddressDTO();
        $addressDTO->city = "{$input['city']} from custom model transformer";

        return $addressDTO;
    }
}
```

We can see that we are instanciating an `AddressDTO`, when extracting code, we don't extract that this statement needs the `use App\AddressDTO;`, we should do it to avoid issues when using CustomTransformers.

On how to do it, I think we need to make it in 3 steps:
- Get all `use` statements in the file
- Extract the `transform` method code
- Check within the extracted code if we need anything that was in the `use` statements

nikophil:
fixes #32

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [GH#10](https://github.com/jolicode/automapper/pull/10) Introduce custom transformers
 - [GH#26](https://github.com/jolicode/automapper/pull/26) Fix mappings involving DateTimeInterface type
 
+### Fixed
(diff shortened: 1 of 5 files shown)


Korbeil (approved):
Amazing ! Thanks for the fix :rocket:

PR #37, opened 2024-01-18
minor: add phpDoc in mappers

nikophil:
This PR adds some phpdoc in generated mappers, which could be useful when debugging:

example:
```php
    /** @param array $value */   <== phpdoc added here
    public function &map($value, array $context = array()) : mixed
    {
        if (null === $value) {
            return $value;
        }
        /** @var \AutoMapper\Tests\Fixtures\AddressDTOSecondReadonlyClass $result */    <== and here
        $result = $context['target_to_populate'] ?? null;
       
        // ...
    }
```


Korbeil (approved):
LGTM, could you add a quick note in CHANGELOG and fix the coding style issue ? :pray:

Korbeil:
Thanks for your contribution @nikophil ! :pray:

PR #38, opened 2024-01-29
refactor: defer transformer resolving step

nikophil:
Here is some more code on the road to use attributes to configure the automapper.

it is still a "Work In Progress" stuff: I focus on keeping phpunit green. I don't want to fight with PHPStan for now, since a lot of things will be refactored (and we're targetting a temporary branch).

About the work in this PR: I feel like we should use `PropertyMapping` objects to resolve the related attributes (see `PropertyMapping::getRelatedAttribute()`), so I needed to remove the `$transformer` from the `PropertyMapping` and resolve it later

### DIFF
--- a/src/Attribute/PropertyAttribute.php
+++ b/src/Attribute/PropertyAttribute.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AutoMapper\Attribute;
+
+use AutoMapper\Extractor\PropertyMapping;
+
+interface PropertyAttribute
+{
+    public function supports(PropertyMapping $propertyMapping): bool;
+}
--- a/src/AutoMapper.php
+++ b/src/AutoMapper.php
@@ -15,8 +15,14 @@ use AutoMapper\Extractor\MapToContextPropertyInfoExtractorDecorator;
 use AutoMapper\Extractor\SourceTargetMappingExtractor;
 use AutoMapper\Generator\MapperGenerator;
 use AutoMapper\Generator\Shared\ClassDiscriminatorResolver;
+use AutoMapper\Generator\TransformerResolver\ChainTransformerResolver;
 use AutoMapper\Loader\ClassLoaderInterface;
 use AutoMapper\Loader\EvalLoader;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataFactory;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataFactoryInterface;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataInterface;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataRegistry;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataRegistryInterface;
 use AutoMapper\Transformer\ArrayTransformerFactory;
 use AutoMapper\Transformer\BuiltinTransformerFactory;
 use AutoMapper\Transformer\ChainTransformerFactory;
@@ -53,28 +59,21 @@ class AutoMapper implements AutoMapperInterface, AutoMapperRegistryInterface, Ma
     public const RELEASE_VERSION = 0;
     public const EXTRA_VERSION = 'DEV';
 
-    /** @var MapperGeneratorMetadataInterface[] */
-    private array $metadata = [];
-
     /** @var GeneratedMapper[] */
     private array $mapperRegistry = [];
 
     public function __construct(
         private readonly ClassLoaderInterface $classLoader,
         private readonly ChainTransformerFactory $chainTransformerFactory,
-        public readonly CustomTransformersRegistry $customTransformersRegistry,
-        private readonly ?MapperGeneratorMetadataFactoryInterface $mapperConfigurationFactory = null
+        private readonly CustomTransformersRegistry $customTransformersRegistry,
+        private readonly MapperGeneratorMetadataRegistryInterface $mapperGeneratorMetadataRegistry,
+        private readonly MapperGenerator $mapperGenerator,
     ) {
-    }
-
-    public function register(MapperGeneratorMetadataInterface $configuration): void
-    {
-        $this->metadata[$configuration->getSource()][$configuration->getTarget()] = $configuration;
     }
 
     public function getMapper(string $source, string $target): MapperInterface
     {
-        $metadata = $this->getMetadata($source, $target);
+        $metadata = $this->mapperGeneratorMetadataRegistry->getMetadata($source, $target);
 
         if (null === $metadata) {
             throw new NoMappingFoundException('No mapping found for source ' . $source . ' and target ' . $target);
@@ -87,7 +86,7 @@ class AutoMapper implements AutoMapperInterface, AutoMapperRegistryInterface, Ma
         }
 
         if (!class_exists($className)) {
-            $this->classLoader->loadClass($metadata);
+            $this->classLoader->loadClass($this->mapperGenerator, $metadata);
         }
 
         $this->mapperRegistry[$className] = new $className();
@@ -102,7 +101,7 @@ class AutoMapper implements AutoMapperInterface, AutoMapperRegistryInterface, Ma
 
     public function hasMapper(string $source, string $target): bool
     {
-        return null !== $this->getMetadata($source, $target);
+        return null !== $this->mapperGeneratorMetadataRegistry->getMetadata($source, $target);
     }
 
     public function map(array|object $source, string|array|object $target, array $context = []): null|array|object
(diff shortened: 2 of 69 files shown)


src/Attribute/PropertyAttribute.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AutoMapper\Attribute;
+
+use AutoMapper\Extractor\PropertyMapping;
+
+interface PropertyAttribute
```
nikophil, line 9: not sure about this name... but it would be nice to share an interface between all attributes we're exposing

src/AutoMapper.php
```diff
@@ -53,28 +59,21 @@ class AutoMapper implements AutoMapperInterface, AutoMapperRegistryInterface, Ma
...
-    private array $metadata = [];
-
     /** @var GeneratedMapper[] */
     private array $mapperRegistry = [];
 
     public function __construct(
         private readonly ClassLoaderInterface $classLoader,
         private readonly ChainTransformerFactory $chainTransformerFactory,
-        public readonly CustomTransformersRegistry $customTransformersRegistry,
-        private readonly ?MapperGeneratorMetadataFactoryInterface $mapperConfigurationFactory = null
```
nikophil, line 66: do you know why the `MapperGeneratorMetadataFactoryInterface` dependency used to be optional? I can't see any case where it is justified

src/AutoMapper.php
```diff
@@ -157,6 +143,16 @@ public function bindCustomTransformer(CustomTransformerInterface $customTransfor
...
 
+    public function register(MapperGeneratorMetadataInterface $configuration): void
+    {
+        $this->mapperGeneratorMetadataRegistry->register($configuration);
+    }
+
+    public function getMetadata(string $source, string $target): ?MapperGeneratorMetadataInterface
+    {
+        return $this->mapperGeneratorMetadataRegistry->getMetadata($source, $target);
+    }
```
nikophil, line 154: I've refactored a bit, in order to introduce a `MapperGeneratorMetadataRegistry` class, which will be useful to de-complexify this class as well as `MapperMetadata` class

src/Loader/ClassLoaderInterface.php
```diff
@@ -13,5 +14,5 @@
  */
 interface ClassLoaderInterface
 {
-    public function loadClass(MapperGeneratorMetadataInterface $mapperMetadata): void;
+    public function loadClass(MapperGenerator $generator, MapperGeneratorMetadataInterface $mapperMetadata): void;
```
nikophil, line 17: I refactored this method, in order to remove the `MapperGenerator` from `EvalLoader` / `FileLoader`'s constructor.
This is really easier to refactor then, because, we don't need to create a full `MapperGenerator` [in the tests](https://github.com/nikophil/automapper/blob/feat/deffer-resolve-transformer/tests/AutoMapperBaseTest.php#L36)

src/AutoMapper.php
```diff
@@ -53,28 +59,21 @@ class AutoMapper implements AutoMapperInterface, AutoMapperRegistryInterface, Ma
...
     private array $mapperRegistry = [];
 
     public function __construct(
         private readonly ClassLoaderInterface $classLoader,
         private readonly ChainTransformerFactory $chainTransformerFactory,
-        public readonly CustomTransformersRegistry $customTransformersRegistry,
-        private readonly ?MapperGeneratorMetadataFactoryInterface $mapperConfigurationFactory = null
+        private readonly CustomTransformersRegistry $customTransformersRegistry,
+        private readonly MapperGeneratorMetadataRegistryInterface $mapperGeneratorMetadataRegistry,
+        private readonly MapperGenerator $mapperGenerator,
```
nikophil, line 70: I'm just realizing this change might not be clever: this forces to instantiate `MapperGenerator` everytime the automapper is used, but as soon as all mappers are generated, nothing regarding "code generation" should be instantiated...

I've done it this way because, we now have a lot more classes to instantiate, and some tests are very cumbersome to update sometimes now...

I think I'll re-work this whole thing. But for now, let's merge it like this (the target branch is a temporary one).

src/Extractor/FromSourceMappingExtractor.php
```diff
@@ -61,35 +61,11 @@ public function getPropertiesMapping(MapperGeneratorMetadataInterface $mapperMet
                 continue;
             }
 
-            $sourceTypes = $this->propertyInfoExtractor->getTypes($mapperMetadata->getSource(), $property);
```
nikophil, line 64: this code (and their equivalent in `FromTargetMappingExtractor` / `SourceTargetMappingExtractor`) have been moved in a `TransformerResolver`

src/Generator/PropertyStatementsGenerator.php
```diff
@@ -38,12 +39,14 @@ public function generate(PropertyMapping $propertyMapping): array
 
         $fieldValueVariable = $variableRegistry->getFieldValueVariable($propertyMapping);
 
-        if ($propertyMapping->hasCustomTransformer()) {
-            $output = $this->customTransformerExtractor->extract($propertyMapping->transformer, $fieldValueVariable, $variableRegistry->getSourceInput());
+        $transformer = $propertyMapping->getTransformer();
+
+        if (is_string($transformer)) {
```
nikophil, line 44: this is kinda "fragile" and gonna be refactored at some point (I'm wondering if I cannot make "custom transformers extractor" implement the main `TranformeInterface`)

src/Generator/Shared/CircularReferenceChecker.php
```diff
@@ -0,0 +1,72 @@
...
+
+declare(strict_types=1);
+
+namespace AutoMapper\Generator\Shared;
+
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataInterface;
+use AutoMapper\MapperMetadata\MapperGeneratorMetadataRegistryInterface;
+use AutoMapper\Transformer\DependentTransformerInterface;
+
+final readonly class CircularReferenceChecker
```
nikophil, line 11: this have been extracted from `MapperMetadata` 

I'd like to de-complexify some objects, this is a part of it

src/Generator/TransformerResolver/ChainTransformerResolver.php
```diff
@@ -0,0 +1,60 @@
...
+use AutoMapper\Attribute\MapTo;
+use AutoMapper\CustomTransformer\CustomTransformerGenerator;
+use AutoMapper\CustomTransformer\CustomTransformersRegistry;
+use AutoMapper\Extractor\PropertyMapping;
+use AutoMapper\Transformer\CallbackTransformer;
+use AutoMapper\Transformer\TransformerFactoryInterface;
+use AutoMapper\Transformer\TransformerInterface;
+use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
+
+final readonly class ChainTransformerResolver implements TransformerResolverInterface
```
nikophil, line 17: I need to think a little bit more about this one, the inner logic might not stay this way

src/Generator/TransformerResolver/FromSourceTransformerResolver.php
```diff
@@ -0,0 +1,79 @@
...
+namespace AutoMapper\Generator\TransformerResolver;
+
+use AutoMapper\CustomTransformer\CustomTransformersRegistry;
+use AutoMapper\Extractor\PropertyMapping;
+use AutoMapper\Transformer\TransformerFactoryInterface;
+use AutoMapper\Transformer\TransformerInterface;
+use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+final readonly class FromSourceTransformerResolver implements TransformerResolverInterface
```
nikophil, line 14: there is no new code here, only extracted from extractors

src/MapperMetadata/MapperMetadata.php
```diff
@@ -241,62 +237,41 @@ private function buildPropertyMapping(): array
         $propertiesMapping = [];
 
         foreach ($this->mappingExtractor->getPropertiesMapping($this) as $propertyMapping) {
+            $transformer = $this->mappingExtractor->resolveTransformer($propertyMapping);
+
+            if (!$transformer) {
+                continue;
+            }
+
+            $propertyMapping->setTransformer($transformer);
```
nikophil, line 246: now we've defered the "transformer resolving step", here is how we decide a "property mapping" should be kept.

src/MapperMetadata/MapperMetadata.php
```diff
@@ -348,4 +323,13 @@ public function getVariableRegistry(): VariableRegistry
...
     }
+
+    public function mapperType(): MapperType
+    {
+        return match(true){
+            'array' === $this->target || 'stdClass' === $this->target => MapperType::FROM_SOURCE,
+            'array' === $this->source || 'stdClass' === $this->source => MapperType::FROM_TARGET,
+            default => MapperType::SOURCE_TARGET
+        };
+    }
```
nikophil, line 334: I think we should use this new enum everywhere in the code

src/Loader/FileLoader.php
```diff
@@ -5,29 +5,30 @@
...
     public function __construct(
-        private readonly MapperGenerator $generator,
         private readonly string $directory,
         private readonly bool $hotReload = true,
     ) {
         $this->printer = new Standard();
     }
 
-    public function loadClass(MapperGeneratorMetadataInterface $mapperMetadata): void
+    public function loadClass(MapperGenerator $mapperGenerator, MapperGeneratorMetadataInterface $mapperMetadata): void
```
nikophil, line 31: I refactored this method, in order to remove the `MapperGenerator` from `EvalLoader` / `FileLoader`'s constructor.
This is really easier to refactor then, because, we don't need to create a full `MapperGenerator` [in the tests](https://github.com/nikophil/automapper/blob/feat/deffer-resolve-transformer/tests/AutoMapperBaseTest.php#L36)

PR #44, opened 2024-03-08
feat(generator): allow skipping uninitialized property when skipping null values

### LINKED ISSUE #43: Handle non-initialized properties
When generating a Mapper, we will have stuff like this: `\AutoMapper\MapperContext::isAllowedAttribute($context, 'scoreOnCreation', $fieldvalue_20 = $value->scoreOnCreation)`.
But if the field is not initialized before it will fail violently. We should handle non-initiliazed properties a better way, `isset` can tell if a property is initialized or not.

joelwurtz:
Fixes #43 

Instead of getting the value we do a check on whether the value is null or undefined, this allow skipping mapping uninitialized property also.

Without this change it php would throw an error as we try to get a uninitialized property value

### DIFF
--- a/src/Extractor/ReadAccessor.php
+++ b/src/Extractor/ReadAccessor.php
@@ -154,6 +154,64 @@ final class ReadAccessor
         throw new CompileException('Invalid accessor for read expression');
     }
 
+    public function getIsNullExpression(Expr\Variable $input): Expr
+    {
+        if (self::TYPE_METHOD === $this->type) {
+            $methodCallExpr = $this->getExpression($input);
+
+            /*
+             * null !== $methodCallExpr
+             */
+            return new Expr\BinaryOp\Identical(
+                new Expr\ConstFetch(new Name('null')),
+                $methodCallExpr,
+            );
+        }
+
+        if (self::TYPE_PROPERTY === $this->type) {
+            if ($this->private) {
+                /*
+                 * When the property is private we use the extract callback that can read this value
+                 *
+                 * @see \AutoMapper\Extractor\ReadAccessor::getExtractIsNullCallback()
+                 *
+                 * $this->extractIsNullCallbacks['property_name']($input)
+                 */
+                return new Expr\FuncCall(
+                    new Expr\ArrayDimFetch(new Expr\PropertyFetch(new Expr\Variable('this'), 'extractIsNullCallbacks'), new Scalar\String_($this->accessor)),
+                    [
+                        new Arg($input),
+                    ]
+                );
+            }
+
+            /*
+             * Use the property fetch to read the value
+             *
+             * isset($input->property_name)
+             */
+            return new Expr\Isset_([new Expr\PropertyFetch($input, $this->accessor)]);
+        }
+
+        if (self::TYPE_ARRAY_DIMENSION === $this->type) {
+            /*
+             * Use the array dim fetch to read the value
+             *
+             * isset($input['property_name'])
+             */
+            return new Expr\Isset_([new Expr\ArrayDimFetch($input, new Scalar\String_($this->accessor))]);
+        }
+
+        if (self::TYPE_SOURCE === $this->type) {
+            return new Expr\BinaryOp\Identical(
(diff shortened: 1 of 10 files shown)


src/MapperContext.php
```diff
@@ -175,9 +175,9 @@ public static function withReference(array $context, string $reference, &$object
     /**
      * Check whether an attribute is allowed to be mapped.
      */
-    public static function isAllowedAttribute(array $context, string $attribute, $value): bool
+    public static function isAllowedAttribute(array $context, string $attribute, bool $valueIsNotNullOrNotUndefined): bool
```
joelwurtz, line 178: Does this method is internal or not ? 

If yes then this is a BC break and i will update the PR accordingly

Korbeil: I would consider this internal, also we included AutoMapper version into Mapper hash, so if you update AutoMapper the Mapper will be re-generated so there won't be any issue there ~

Korbeil (approved):
LGTM :+1:

src/MapperContext.php
nikophil: at some point, shouldn't all public methods here be marked internal?

src/Extractor/ReadAccessor.php
```diff
@@ -193,4 +251,36 @@ public function getExtractCallback(string $className): ?Expr
             new Arg(new Scalar\String_($className)),
         ]);
     }
+
+    /**
+     * Get AST expression for binding closure when dealing with a private property.
+     */
+    public function getExtractIsNullCallback(string $className): ?Expr
+    {
+        if ($this->type !== self::TYPE_PROPERTY || !$this->private) {
```
nikophil, line 260: maybe we can extract a private method like `needsExtractIsNullCallback()` used here and in `getIsNullExpression()`? this would make explicit that these cases are linked

src/Generator/PropertyConditionsGenerator.php
```diff
@@ -112,16 +112,10 @@ private function isAllowedAttribute(PropertyMapping $propertyMapping): ?Expr
...
-        $sourcePropertyAccessor = new Expr\Assign(
-            $variableRegistry->getFieldValueVariable($propertyMapping),
-            $propertyMapping->readAccessor->getExpression($variableRegistry->getSourceInput())
-        );
-
         return new Expr\StaticCall(new Name\FullyQualified(MapperContext::class), 'isAllowedAttribute', [
             new Arg($variableRegistry->getContext()),
             new Arg(new Scalar\String_($propertyMapping->property)),
-            new Arg($sourcePropertyAccessor),
+            new Arg($propertyMapping->readAccessor->getIsNullExpression($variableRegistry->getSourceInput())),
```
nikophil, line 118: this might change the generated code: maybe the phpdoc of this method should be updated?

nikophil (approved):
LGTM :+1:

src/Extractor/ReadAccessor.php
```diff
@@ -154,6 +154,64 @@ public function getExpression(Expr\Variable $input): Expr
         throw new CompileException('Invalid accessor for read expression');
     }
 
+    public function getIsNullExpression(Expr\Variable $input): Expr
```
nikophil, line 157: shouldn't this method be called `getIsNotNullExpression()`?

src/MapperContext.php
joelwurtz: I will mark the class as internal then

src/Extractor/ReadAccessor.php
joelwurtz: On the other method it's just a sub case, i think it makes it more abstract for little gain will keep it like this

src/Extractor/ReadAccessor.php
joelwurtz: Nope, it really give a check whether the value is null `isset($value) | null === $source->getFoo()` and not the contrary

src/MapperContext.php
joelwurtz: I mark the method as internal

PR #45, opened 2024-03-08
chore(cs): fix composer.lock for phpstan and php cs fixer

joelwurtz:
It also fixes cs regard the installed cs fixer version

This should avoid having CI unstable if php-cs-fixer change in the future


Korbeil (approved):
Thanks for the contribution @joelwurtz !

Issue #4, opened 2023-09-22
ObjectTransformer is not available when custom MapperConfigurationInterface is processed


Korbeil:
Original issue: https://github.com/janephp/janephp/issues/745
Reported by: @weaverryan

**Jane version(s) affected**: 7.5.3

**Description**  
Hi!

In the AutoMapper bundle, if you create a custom `MapperConfigurationInterface` and call `$metadata->getPropertiesMapping()`, the mappings will be generated BEFORE the `ObjectTransformer` was loaded into the system. The problem comes with how the objects are instantiated:

A) When the container creates the`Jane\Component\AutoMapper\Transformer\ChainTransformerFactory` service, in the compiled container, it eventually calls:

```php
$instance->addTransformerFactory(new \Jane\Component\AutoMapper\Transformer\ObjectTransformerFactory(($container->services['Jane\\Component\\AutoMapper\\AutoMapperInterface'] ?? $container->load('getAutoMapperInterfaceService'))));
```

In order to create the `ObjectTransformerFactory`, it needs to load the `AutoMapperInterface` service so it can be passed to it.

B) But, when the `AutoMapperInterface` is being created, in the cached container, it will call:

```php
$instance->addMapperConfiguration(new \App\Automapper\UserToUserApiMapperConfiguration());
```

(where `UserToUserApiMapperConfiguration` is the custom mapper configuration class).

C) The `AutoMapper::addMapperConfiguration()` method - https://github.com/janephp/janephp/blob/next/src/Bundle/AutoMapperBundle/AutoMapper.php#L13 - causes the `process()` method to be called on the custom mapper configuration. However, at this point, the `ObjectTransformerFactory` hasn't yet finished being added to the system.

This is a minor issue - it can just cause WTF moments. I couldn't figure out why a property was NOT apparently being mapped... but then WOULD be mapped in the end.

**How to reproduce**  

```php
symfony new reproducing_jane_missing_object_transformer
cd reproducing_jane_missing_object_transformer
composer require jane-php/automapper-bundle
```

Create a mapping config class - throw an exception:

```php
<?php

namespace App;

use Jane\Bundle\AutoMapperBundle\Configuration\MapperConfigurationInterface;
use Jane\Component\AutoMapper\MapperGeneratorMetadataInterface;

class TestingMapperConfiguration implements MapperConfigurationInterface
{
    public function process(MapperGeneratorMetadataInterface $metadata): void
    {
        throw new \Exception();
    }

    public function getSource(): string
    {
[...]

PR #48, opened 2024-03-09
fix(transformer): change the way transformer factory are injected to make it work as soon as automapper is created

joelwurtz:
This fix https://github.com/jolicode/automapper/issues/4

Mainly the issue is we don't know in which order the process / addTransformerFactory will be done in the DIC of Symfony.

To havoid doing something complicated in the bundle i change the way factories are created.

So now they can be created before the automapper, and registry / chain transformer will be set when they are injected.

This also allow only implementating an interface to have the registry / chain set and will simplify DIC configuration in the bundle

Now when we do `new AutoMapper()` it should be ready instantly


src/AutoMapperRegistryAwareInterface.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace AutoMapper;
+
+/**
+ * Allows to use a AutoMapperRegistry.
+ *
+ * @author Joel Wurtz <jwurtz@jolicode.com>
+ */
+interface AutoMapperRegistryAwareInterface
```
Korbeil, line 12: should it be tagged as internal ?

src/AutoMapper.php
```diff
@@ -145,6 +146,9 @@ public function getMetadata(string $source, string $target): ?MapperGeneratorMet
         return $this->metadata[$source][$target];
     }
 
+    /**
+     * @deprecated since 8.1, will be removed in 9.0.
+     */
```
Korbeil, line 151: Would require a `trigger_deprecation` call to inform people to not use it anymore and what is the replacement.

src/Transformer/ChainTransformerFactory.php
```diff
@@ -4,43 +4,90 @@
...
+                $factory->setAutoMapperRegistry($autoMapperRegistry);
+            }
+        }
+    }
 
     /**
      * Biggest priority is MultipleTransformerFactory with 128, so default priority will be bigger in order to
      * be used before it, 256 should be enough.
+     *
+     * @deprecated since 8.1, will be removed in 9.0. Pass the factory into the constructor instead
```
Korbeil, line 47: Requires a `trigger_deprecation` call

src/Transformer/ChainTransformerFactory.php
```diff
@@ -4,43 +4,90 @@
...
         }
-        $this->factories[$priority][] = $transformerFactory;
+
+        $this->factories[] = $transformerFactory;
+        $this->sortFactories();
     }
 
+    /**
+     * @deprecated since 8.1, will be removed in 9.0.
+     */
```
Korbeil, line 84: Requires a `trigger_deprecation` call

Korbeil (approved):
LGTM :+1: 

Thanks for the new feature !
Some comments mostly about deprecation warninigs.

nikophil (approved):
LGTM

PR #54, opened 2024-03-11
feat: introduce MapperContext::DATETIME_FORCE_TIMEZONE

nikophil:


src/Transformer/StringToDateTimeTransformer.php
```diff
@@ -36,7 +36,13 @@ public function transform(Expr $input, Expr $target, PropertyMapping $propertyMa
         /*
          * Create a \DateTime[Immutable] object from a string.
          *
-         * \DateTimeImmutable::createFromFormat($context[MapperContext::DATETIME_FORMAT] ?? \DateTimeInterface::RFC3339, $input);
+         * ```php
+         * \DateTimeImmutable::createFromFormat(
+         *      $context[MapperContext::DATETIME_FORMAT] ?? \DateTimeInterface::RFC3339,
+         *      $input,
+         *      isset($context[MapperContext::DATETIME_FORCE_TIMEZONE]) ? new \DateTimeZone($context[MapperContext::DATETIME_FORCE_TIMEZONE]) : null
```
nikophil, line 43: or I can add a static method `MapperContext::getForcedTimeZone($context)`

this way, we could handle the error if the passed timezone is invalid, and make it explicit that it comes from automapper's context. WDYT?

src/MapperContext.php
```diff
@@ -246,4 +247,19 @@ public static function withNewContext(array $context, string $attribute): array
 
         return $context;
     }
+
+    public static function getForcedTimezone(array $context): ?\DateTimeZone
```
Korbeil, line 251: I think you should put:
```suggestion
 /**
     * @param array{datetime_force_timezone?: string} $context
     */
    public static function getForcedTimezone(array $context): ?\DateTimeZone
```

Korbeil (approved):

Issue #28, opened 2024-01-02
[DX] Normalize data directly thanks to php attributes


nikophil:
I think what is done by valinor with attributes is pretty nice:

https://twitter.com/Rommsteinz/status/1739673469724299392?t=5oVgUTC1atb1UacsHkB_bw

We should be able to leverage some concept like this in automapper

Issue #16, opened 2023-10-31
[Feature] introduce attribute `#[AutoMapper\Ignore]`


nikophil:
presently, the only way to ignore a property is to use serializer's `#[Ignore]` attribute (or to use automapper context's `IGNORED_ATTRIBUTES`, but this forces to rely on properties names, and cannot be used in as a generic mechanism)

Myabe we could then introduce its own ignore attribute for the automapper?

joelwurtz:
Fix by https://github.com/jolicode/automapper/pull/59

PR #59, opened 2024-03-14
feat(attribute): add map to attribute

joelwurtz:
Add `#[MapTo]` attribute which allows to : 

 * Add a mapping on a previously not found property
 * Ignore a property during a Map ref #16 
 * Customize the target property ref #34 
 * Allow to use a custom transformer : callable / transformer service ref #28

### DIFF
--- a/src/EventListener/MapToListener.php
+++ b/src/EventListener/MapToListener.php
@@ -9,6 +9,7 @@ use AutoMapper\Event\GenerateMapperEvent;
 use AutoMapper\Event\PropertyMetadataEvent;
 use AutoMapper\Event\SourcePropertyMetadata;
 use AutoMapper\Event\TargetPropertyMetadata;
+use AutoMapper\Exception\BadMapDefinitionException;
 use AutoMapper\Transformer\CallableTransformer;
 use AutoMapper\Transformer\CustomTransformer\CustomModelTransformer;
 use AutoMapper\Transformer\CustomTransformer\CustomModelTransformerInterface;
@@ -76,7 +77,7 @@ final readonly class MapToListener
                 );
 
                 if (isset($event->properties[$property->target->name])) {
-                    // @TODO throw exception
+                    throw new BadMapDefinitionException(sprintf('There is already a MapTo attribute with target "%s" in class "%s".', $property->target->name, $event->mapperMetadata->source));
                 }
 
                 $event->properties[$property->target->name] = $property;
--- a/src/EventListener/MapToListener.php
+++ b/src/EventListener/MapToListener.php
@@ -19,8 +19,9 @@ use AutoMapper\Transformer\CustomTransformer\CustomTransformersRegistry;
 
 final readonly class MapToListener
 {
-    public function __construct(private CustomTransformersRegistry $customTransformersRegistry)
-    {
+    public function __construct(
+        private CustomTransformersRegistry $customTransformersRegistry
+    ) {
     }
 
     public function __invoke(GenerateMapperEvent $event): void
@@ -34,7 +35,7 @@ final readonly class MapToListener
         foreach ($properties as $reflectionProperty) {
             $mapToAttributes = $reflectionProperty->getAttributes(MapTo::class);
 
-            if (empty($mapToAttributes)) {
+            if (0 === \count($mapToAttributes)) {
                 continue;
             }
 
@@ -76,7 +77,7 @@ final readonly class MapToListener
                     $mapToAttributeInstance->ignore,
                 );
 
-                if (isset($event->properties[$property->target->name])) {
+                if (\array_key_exists($property->target->name, $event->properties)) {
                     throw new BadMapDefinitionException(sprintf('There is already a MapTo attribute with target "%s" in class "%s".', $property->target->name, $event->mapperMetadata->source));
                 }
 
--- a/src/EventListener/MapToListener.php
+++ b/src/EventListener/MapToListener.php
@@ -9,10 +9,16 @@ use AutoMapper\Event\GenerateMapperEvent;
 use AutoMapper\Event\PropertyMetadataEvent;
 use AutoMapper\Event\SourcePropertyMetadata;
 use AutoMapper\Event\TargetPropertyMetadata;
+use AutoMapper\Transformer\CallableTransformer;
+use AutoMapper\Transformer\CustomTransformer\CustomModelTransformer;
+use AutoMapper\Transformer\CustomTransformer\CustomModelTransformerInterface;
+use AutoMapper\Transformer\CustomTransformer\CustomPropertyTransformer;
+use AutoMapper\Transformer\CustomTransformer\CustomPropertyTransformerInterface;
+use AutoMapper\Transformer\CustomTransformer\CustomTransformersRegistry;
 
-class MapToMapperListener
+final readonly class MapToListener
 {
-    public function __construct()
+    public function __construct(private CustomTransformersRegistry $customTransformersRegistry)
     {
     }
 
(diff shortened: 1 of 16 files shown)


src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,87 @@
...
+use AutoMapper\Transformer\CustomTransformer\CustomModelTransformerInterface;
+use AutoMapper\Transformer\CustomTransformer\CustomPropertyTransformer;
+use AutoMapper\Transformer\CustomTransformer\CustomPropertyTransformerInterface;
+use AutoMapper\Transformer\CustomTransformer\CustomTransformersRegistry;
+
+final readonly class MapToListener
+{
+    public function __construct(private CustomTransformersRegistry $customTransformersRegistry)
+    {
+    }
```
Korbeil, line 24:
```suggestion
    public function __construct(
        private CustomTransformersRegistry $customTransformersRegistry
    ) {
    }
```

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,87 @@
...
+        if ($event->mapperMetadata->sourceReflectionClass === null) {
+            return;
+        }
+
+        $properties = $event->mapperMetadata->sourceReflectionClass->getProperties();
+
+        foreach ($properties as $reflectionProperty) {
+            $mapToAttributes = $reflectionProperty->getAttributes(MapTo::class);
+
+            if (empty($mapToAttributes)) {
```
Korbeil, line 37:
```suggestion
            if (0 === \count($mapToAttributes)) {
```

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,87 @@
...
+                continue;
+            }
+
+            foreach ($mapToAttributes as $mapToAttribute) {
```
Korbeil, line 41: Do we have some kind of priority between attributes there ? Or it's first written onto the class that is first here ?

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,87 @@
...
+
+                    if (\is_string($mapToAttributeInstance->transformer) && $customTransformer = $this->customTransformersRegistry->getCustomTransformer($mapToAttributeInstance->transformer)) {
+                        if ($customTransformer instanceof CustomModelTransformerInterface) {
+                            $transformer = new CustomModelTransformer($mapToAttributeInstance->transformer);
+                        }
+
+                        if ($customTransformer instanceof CustomPropertyTransformerInterface) {
+                            $transformer = new CustomPropertyTransformer($mapToAttributeInstance->transformer);
+                        }
+                    } elseif (@\is_callable($mapToAttributeInstance->transformer, false, $callableName) && $callableName !== null) {
```
Korbeil, line 65: Is that `@` required ?

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,87 @@
...
+                $property = new PropertyMetadataEvent(
+                    $event->mapperMetadata,
+                    $sourceProperty,
+                    $targetProperty,
+                    $mapToAttributeInstance->maxDepth,
+                    $transformer,
+                    $mapToAttributeInstance->ignore,
+                );
+
+                if (isset($event->properties[$property->target->name])) {
```
Korbeil, line 79:
```suggestion
                if (array_key_eixsts($property->target->name, $event->properties)) {
```

src/Transformer/CallableTransformer.php
```diff
@@ -0,0 +1,30 @@
...
+use AutoMapper\Generator\UniqueVariableScope;
+use AutoMapper\Metadata\PropertyMetadata;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar;
+
+class CallableTransformer implements TransformerInterface
+{
+    public function __construct(private string $callable)
+    {
```
Korbeil, line 16:
```suggestion
    public function __construct(
        private string $callable
    ) {
```

src/Transformer/CallableTransformer.php
```diff
@@ -0,0 +1,30 @@
...
+
+use AutoMapper\Generator\UniqueVariableScope;
+use AutoMapper\Metadata\PropertyMetadata;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar;
+
+class CallableTransformer implements TransformerInterface
+{
+    public function __construct(private string $callable)
```
Korbeil, line 15: Is that because we can't handle "real" callable in attributes ? (things like `fn($var) => $var` for exemple ?

src/EventListener/MapToListener.php
Korbeil: No reason to have priority since we throw an exception when we have a rule for the same target.

src/Transformer/CallableTransformer.php
joelwurtz: Yes otherwise it's "impossible" to generate (we can use the previous callback extractor but imo this is a rabbithole)

Korbeil (approved):
LGTM

src/EventListener/MapToListener.php
joelwurtz: Not really but current support is not great, we should do like the callback validator of symfony in term of support without the `Closure`

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,260 @@
...
+
+    private function setPropertiesFromSource(GenerateMapperEvent $event): void
+    {
+        if ($event->mapperMetadata->sourceReflectionClass === null) {
+            return;
+        }
+
+        $properties = $event->mapperMetadata->sourceReflectionClass->getProperties();
+
+        foreach ($properties as $reflectionProperty) {
```
nikophil, line 51: I think this method (and `setPropertiesFromTarget()`) can be simplified by looping over
`[...$properties, ...$methods]`

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,260 @@
...
+
+        $this->setPropertiesFromSource($event);
+        $this->setPropertiesFromTarget($event);
+    }
+
+    private function setPropertiesFromSource(GenerateMapperEvent $event): void
+    {
+        if ($event->mapperMetadata->sourceReflectionClass === null) {
+            return;
+        }
```
nikophil, line 47: seems irrelevant, because this condition is already in `__invoke()` method

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    public function __invoke(GenerateMapperEvent $event): void
+    {
+        if ($event->mapperMetadata->sourceReflectionClass === null) {
+            return;
+        }
+
+        $this->setPropertiesFromSource($event);
+        $this->setPropertiesFromTarget($event);
```
nikophil, line 40: shouldn't this `MapToListener` only handle `MapTo` attributes? and handle `MapFrom` in another listener?
Here, because of the previous condition, we cannot deal with "array to object" transformations (I don't know if you are planning to do this in another PR)

src/EventListener/MapToListener.php
```diff
@@ -0,0 +1,260 @@
...
+        private CustomTransformersRegistry $customTransformersRegistry,
+        private InflectorInterface $inflector = new EnglishInflector(),
+    ) {
+    }
+
+    public function __invoke(GenerateMapperEvent $event): void
+    {
+        if ($event->mapperMetadata->sourceReflectionClass === null) {
+            return;
+        }
```
nikophil, line 37: here is an idea we could use all over the code to improve readability: I think we should introduce an enum which would looks like:
```php
enum MappingType
{
    case FROM_SOURCE; // (aka NORMALIZATION)
    case FROM_TARGET; // (aka DENORMALIZATION)
    case SOURCE_TARGET; // (aka OBJECTS_MAPPING)
}
```
and rely on this enum instead of relying on this kind of condition which is kinda meaningless

src/Transformer/CallableTransformer.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace AutoMapper\Transformer;
+
+use AutoMapper\Generator\UniqueVariableScope;
+use AutoMapper\Metadata\PropertyMetadata;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar;
+
+class CallableTransformer implements TransformerInterface
```
nikophil, line 13: final / readonly / internal ?

tests/Fixtures/MapTo/Bar.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace AutoMapper\Tests\Fixtures\MapTo;
+
+use AutoMapper\Attribute\MapFrom;
+
+class Bar
+{
+    #[MapFrom(source: 'array', ignore: true)]
```
nikophil, line 11: does this work?

PR #63, opened 2024-03-16
feat(bundle): Merge bundle directly into automapper, will replace automapper-bundle

joelwurtz:


Korbeil (approved):
Thanks for the PR & nice additions to the docs ! :muscle:

PR #67, opened 2024-03-17
feat(lazy): support lazy mapping

joelwurtz:
This introduce the possibility to do lazy mapping

The following call : 

```
$userDto = $this->autoMapper->map($user, Fixtures\UserDTO::class, [MapperContext::LAZY_MAPPING => true]);
```

Will produce a lazy ghost object for userDto which will only be mapped when calling one of its value
Nested objects are also lazy loaded

This will allow to map large complex tree instantly and only map values when it's needed, this may decreses performance in most cases so it's not enabled by default, but in some cases where only a part of a data is needed this may be useful and save process / memory

This also opens the door to json streaming (will need a special LazyArray object but it's totally doable)


joelwurtz:
I put this to draft, want to add some tests and other use cases before completing it

PR #68, opened 2024-03-18
chore(deps): allow to use sf 5.4 for most things, remove useless deps / suggests

joelwurtz:
This allow to make it work with all symfony maintained versions (which should be our goal IMO)


Korbeil (approved):
LGTM

PR #75, opened 2024-03-19
Types: better matching between types to better handle multiple types

joelwurtz:


src/EventListener/MapFromListener.php
```diff
@@ -59,6 +59,7 @@ private function addPropertyFromTarget(GenerateMapperEvent $event, MapFrom $mapF
             $event->mapperMetadata,
             $sourceProperty,
             $targetProperty,
+            null,
             $mapFrom->maxDepth,
             $this->getTransformerFromMapAttribute($event->mapperMetadata->target, $mapFrom),
```
nikophil, line 64: IMO that's where named parameters are useful
```php
$property = new PropertyMetadataEvent(
            mapperMetadata: $event->mapperMetadata,
            source: $sourceProperty,
            target: $targetProperty,
            maxDepth: $mapFrom->maxDepth,
            transformer: $this->getTransformerFromMapAttribute($event->mapperMetadata->target, $mapFrom),
            ignored: $mapFrom->ignore,
        );
```

src/Extractor/MappingExtractorInterface.php
```diff
@@ -22,10 +22,7 @@ interface MappingExtractorInterface
      */
     public function getProperties(string $class): iterable;
 
-    /**
-     * @return array{0: Type[], 1: Type[]}
-     */
-    public function getTypes(string $source, string $sourceProperty, string $target, string $targetProperty): array;
+    public function getTypes(string $source, string $sourceProperty, string $target, string $targetProperty): TypesMatching;
```
nikophil, line 25: this return type is really nicer :grin:

src/Metadata/TypesMatching.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+namespace AutoMapper\Metadata;
+
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * @internal
+ *
+ * @extends \SplObjectStorage<Type, Type[]>
```
nikophil, line 12:
```suggestion
 * @extends \SplObjectStorage<Type, Type>
```
?

src/Transformer/CheckTypeInterface.php
```diff
@@ -0,0 +1,17 @@
...
+namespace AutoMapper\Transformer;
+
+use AutoMapper\Generator\UniqueVariableScope;
+use AutoMapper\Metadata\PropertyMetadata;
+use PhpParser\Node\Expr;
+
+/**
+ * @internal
+ */
+interface CheckTypeInterface
```
nikophil, line 14: maybe add few words on how to use this interface?

src/Metadata/TypesMatching.php
joelwurtz: No it's correct, if you have a field that accept a string or an other type we want to guess the transformer from the first possible type, by example if we have `array` -> `\Traversable|array` as a type we will use `array` as we don't know (yet) how to do a traversable it will be correct since the type allow both

If we want a single relationship we need `[Type, Type][]` this is possible with that but it will change too much current implementation

Issue #5, opened 2023-09-22
bug: problem when getter type is different from property type


nikophil:
**Description**  
when using `#[MapToAttribute]`, when the getter type is different from the property type, the automapper guesses the type from the property, although it should guess it from the getter's return type

**How to reproduce**  
```php
class SomeClass
{
	public function __construct(private SomeEnum $enum){}

    public function getEnum(#[MapToAttribute('foo) string $foo]): string
    {
		return "{$foo}-{$this->enum->value}";
	}
}

$automapper->map(new SomeClass(SomeEnum::FOO, 'array'));
```
this code produces an error, because the generated mapper thinks `SomeClass::getEnum()` will return an enum

**Possible Solution**  
don't know yet, need to dig into the generator

joelwurtz:
Fixed in #77

PR #77, opened 2024-03-19
fix(types): try to get types from read accessor / write mutator first

joelwurtz:
Partially fix https://github.com/jolicode/automapper/issues/5


Korbeil (commented):
Could we add a comment about all the code that should be within Symfony later on ?
Cause it's a lot of code duplication (and it's indeed to fix our issue), but this code shouldn't be on our side in the end.

PR #79, opened 2024-03-19
feat(transformer): refactor the way to create custom transformer

joelwurtz:
I rewamped the way to create `CustomTransformer` as current interfaces were sometimes misleading : 

First it has been renamed to `PropertyTransformer` i think Model was misleading, also the Model was only working when it was a sub property (if called directly it was not working)

There is now 3 interfaces : 

#### PropertyTransformerInterface

```php
/**
 * Transform a property value during a mapping.
 *
 * @experimental
 */
interface PropertyTransformerInterface
{
    /**
     * @param mixed                       $value   the value of the property to transform, can be null if there is no way to read the data from the mapping
     * @param object|array<string, mixed> $source  the source input on which the custom transformation applies
     * @param array<string, mixed>        $context Context during mapping
     */
    public function transform(mixed $value, object|array $source, array $context): mixed;
}
```

This interface is the core, it allow to create a transformer and use it in the `#[MapTo]` and `#[MapFrom]` attribute, there is no more separation between model and property, since it receives both : 

 * `$value` is the value readed from the source property, it may be null when there is no way to read it from source (like a field that need to be mapped from various fields so the source property does not exists)
 * `$source` is the whole source object, this allow to get fields from the source object if needed
 * `$context` is the context being used to map the object, this can be useful when some data may comes from the context (like a `format` property to have different behavior in `json` / `xml` / ...)
 
I was able to do the same behavior as before with the 2 different interfaces with only this class, so i think it's better and clearer for everyone

If latter we want to support model transformation at the root of the Mapper i.e. `map($object, Target::class)` i think it will be better to offer some DX to allow creating its own mapper by implementing `MapperInterface` but it's not the subject here

#### PropertyTransformerSupportInterface

```php
/**
 * Allow to automatically register a transformer for a specific property.
 *
 * @experimental
 */
interface PropertyTransformerSupportInterface
{
    /**
     * When implemented with a PropertyTransformerInterface, this method is called to check if the transformer supports the given properties.
     *
     * This method is not need if the transformer is set in an #[MapTo] or #[MapFrom] attribute
     *
     * @param TypesMatching          $types          The types of the source and target properties
     * @param SourcePropertyMetadata $source         The source property metadata
     * @param TargetPropertyMetadata $target         The target property metadata
     * @param MapperMetadata         $mapperMetadata The mapper metadata
     */
    public function supports(TypesMatching $types, SourcePropertyMetadata $source, TargetPropertyMetadata $target, MapperMetadata $mapperMetadata): bool;
}
```

This interface use the same properties as the transformer factory, which should be more than enough for a majority of use cases, it allows to register a transformer automatically to a mapping if the supports return true.

This may be useful for some third party libraries that want to add their behavior given a class / specific configuration (like an Entity from an ORM or a Resource for an API)

#### PrioritizedPropertyTransformerInterface

```php
/**
 * @experimental
 */
interface PrioritizedPropertyTransformerInterface
{
    public function getPriority(): int;
}
```

The same as before, it allows to order transformers when not using symfony


Korbeil (approved):
LGTM, it looks really better than how it was ! :tada:

PR #80, opened 2024-03-19
feat(if): add if feature to map to / map from attributes

joelwurtz:
There were certainly be edge case, but i think a lot of use cases can already be handled with that

Just wondering if having expression language as a deps is wanted or do we want to make this feature optional ?

Miss documentation, will add it tomorrow


Korbeil (changes requested):
Could you add some documentation about this feature ?

joelwurtz:
I plan to rewamp the documentation very soon, i will write it when doing that

PR #81, opened 2024-03-20
feat(map-to): allow map to / from attribute in class when declaring a transformer and a name

joelwurtz:
This allow to declare a MapTo / MapFrom in the class, useful when there is no relation to a field, in the case transformer will use the source object


Korbeil (commented):
A line or two about this feature for the docs ?

joelwurtz:
Like #80 I plan to rewamp the documentation very soon, i will write it when doing that

PR #90, opened 2024-03-22
chore(test): allow to run symfony app in tests

joelwurtz:


composer.json
```diff
@@ -31,8 +31,12 @@
         "symfony/filesystem": "^5.4 || ^6.0 || ^7.0",
         "symfony/framework-bundle": "*",
         "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0",
+        "symfony/property-access": "^5.4 || ^6.0 || ^7.0",
```
lyrixx, line 34: IMHO, you should use 6.4 everywhere, 6.0..6.3 are EOL

joelwurtz: As long as we support 5.4 there i think any 6 version will work, let's be nice to people that didn't upgrade yet

lyrixx: That's weird. There is no reason to upgrade from 5.4 to 6.0. the correct things to do is 5.4 to 6.4

joelwurtz: More than someone still in 6.1 that want to install this bundle

PR #92, opened 2024-03-22
chore(doc): add doc about mapping

joelwurtz:


docs/getting-started/index.md
```diff
@@ -3,4 +3,4 @@
 - [Understanding the `source` and `target`](source-and-target.md)
 - [Using the context](context.md)
 - [Configuration](configuration.md)
-- [Cache and Loaders](loader.md)
```
Korbeil, line 6: rename the file cache instead of loader ?

PR #96, opened 2024-03-24
feat(mapper): add a way to instantiate object in mapper with providers

joelwurtz:


src/Provider/ProviderRegistry.php
```diff
@@ -0,0 +1,42 @@
...
+declare(strict_types=1);
+
+namespace AutoMapper\Provider;
+
+/**
+ * @internal
+ */
+final readonly class ProviderRegistry
+{
+    /** @var array<string, ProviderInterface> */
```
Korbeil, line 12: Should be `@var array<class-string, ProviderInterface>`

PR #101, opened 2024-03-27
feat(normalize): add a bunch of tests from symfony/serializer component, to verify behavior

joelwurtz:


bench/composer.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "automapper/bench",
```
Korbeil, line 2: I would prefer to split that in another PR please 🙏 (And maybe add a quick note about the output in documentation too ?)

joelwurtz: This is mainly for internal, to check some behavior and keeps them versioned, so we can link to them to why we do that in pull request or issue, not sure if we want that in doc ? could add a readme.md though

Korbeil: Let's add a readme then 👍

joelwurtz: done

PR #103, opened 2024-03-31
feat(debug): add debug command and profiler for symfony

joelwurtz:
This add a debug command that allow to debug a mapper with 

```bash
$ php bin/console debug:mapper 'AutoMapper\Tests\Bundle\Resources\App\Entity\FooMapTo' array
```

Here is a example screenshot and what it output

image

This also add a profiler that show which mappers has been used during a request, this only show mapper if there were generated during the request (that's why i think it should always generate mapper on debug by default see #https://github.com/jolicode/automapper/pull/102 )

Here is some screenshot of the profiler

image

image


Korbeil:
Is it possible to do it without Symfony bundle ?

joelwurtz:
> Is it possible to do it without Symfony bundle ?

It should be, but this is actually available without anything with the bundle, if not using a bundle there is 2 options, either leave it like this and tell the user to inject the command in a `Application` if he has symfony/console component

Or provide a binary to this repository with our own application

Don't want to do that in here, should be another PR

> Could you add an example ?

Can you be more precise ? does the current screenshot not enough here on what it gives ?

Issue #8, opened 2023-10-03
Default values in target are not handled


nikophil:
This [PR](https://github.com/jolicode/automapper/pull/7) made me realize there is another problem, a little bit more complex to mitigate:

```
readonly class SomeClass
{
    public function __construct(public int|null $foo = 1, public int $bar = 0)
    {
    }
}

$autoMapper->map([], SomeClass::class);
```

two problems here:
- `foo` defaults to `null` whereas it should default to `1`
- the above code  creates an error because key `bar` is not present, although the property has a default value

I'm not sure how we could fix this... maybe by adding the default value in the `WriteMutator`?

PR #104, opened 2024-03-31
feat(constructor): use a strategy to have consistent behavior for constructor

joelwurtz:
This fix https://github.com/jolicode/automapper/issues/8

It introduces a new "way" to handle constructor : 

 * auto : will try to use constructor when possible
 * never : same as "allowConstructor: false" as before
 * always : always use the target constructor, this force user to add default value if needed in the context (otherwise il will throw an exception)


src/Configuration.php
```diff
@@ -6,6 +6,10 @@
 
 final readonly class Configuration
 {
+    public const CONSTRUCTOR_STRATEGY_AUTO = 'auto';
+    public const CONSTRUCTOR_STRATEGY_ALWAYS = 'always';
+    public const CONSTRUCTOR_STRATEGY_NEVER = 'never';
```
Korbeil, line 11: I would prefer us using an Enum, what do you think about it ?

src/Generator/CreateTargetStatementsGenerator.php
```diff
@@ -195,21 +170,17 @@ private function constructorArguments(GeneratorMetadata $metadata): array
...
-        }
-
         $variableRegistry = $metadata->variableRegistry;
         $constructVar = $variableRegistry->getVariableWithUniqueName('constructArg');
         $fieldValueExpr = $propertyMetadata->source->accessor?->getExpression($variableRegistry->getSourceInput());
 
         if (null === $fieldValueExpr) {
             if (!($propertyMetadata->transformer instanceof AllowNullValueTransformerInterface)) {
-                return null;
+                return [null, null, null];
```
Korbeil, line 183: A bit cursed, why this change ? `null` would be still fine there no ?

joelwurtz: Don't change really anything, and phpstan works better that way, so i made it like this

src/Generator/CreateTargetStatementsGenerator.php
```diff
@@ -218,6 +189,24 @@ private function constructorArgument(GeneratorMetadata $metadata, PropertyMetada
         /* Get extract and transform statements for this property */
         [$output, $propStatements] = $propertyMetadata->transformer->transform($fieldValueExpr, $constructVar, $propertyMetadata, $variableRegistry->getUniqueVariableScope(), $variableRegistry->getSourceInput());
 
+        $defaultValueExpr = new Expr\Throw_(
```
nikophil, line 192: question: why don't we throw like... now? we know the mapping process will create an error. And it has the advantage to occur sooner, moreover if the mappers generation occur at warmup time

src/Metadata/GeneratorMetadata.php
```diff
@@ -68,10 +69,14 @@ public function isSourceUserDefined(): bool
 
     public function hasConstructor(): bool
     {
-        if (!$this->allowConstructor) {
+        if ($this->constructorStrategy === ConstructorStrategy::NEVER) {
```
nikophil, line 72: in which case would we want to never use the constructor? if the target class has a constructor, wouldn't it break?

docs/bundle/configuration.md
```diff
@@ -9,7 +9,7 @@ Then configure the bundle to your needs, for example:
 ```yaml
 automapper:
   class_prefix: "Symfony_Mapper_"
-  allow_constructor: true
+  constructor_strategy: 'auto'
```
nikophil, line 12: shouldn't we allow to configure this at runtime, thanks to the context?

src/Generator/CreateTargetStatementsGenerator.php
joelwurtz: The mapping will not throw if you provide default constructor argument in the context, there may be case where this is needed

src/Metadata/GeneratorMetadata.php
joelwurtz: Depends on what you use your mapping for, by example doctrine never use the constructor when instantating an entity from the database. At least having this, ensure a consistent behavior there is no "magic"

src/Generator/CreateTargetStatementsGenerator.php
nikophil: ok ineed! I knew I was missing something :sweat_smile:

docs/bundle/configuration.md
joelwurtz: I think having this at runtime would loose performance over a "niche" feature, my goal with that PR is more : I want to generate mapping with the same consistent behavior (never use constructor, always use it, try to use it if possible)

Having this at runtime will complicate thing IMO, let's wait for someone who need that to make this change if needed ?

nikophil (approved):

PR #114, opened 2024-04-22
Introducing Mapper Attribute

joelwurtz:
This PR introduces a Mapper attribute, which will have the following goals : 

 * [x] Allow to override generator options by mapper instead of having some globals (like disable - enable check attributes by mapper)
 * [x] Allow to autoregister mappers having this attribute (we provide a path to the config and all file having this attribute will be auto registered)
 
I introduce only existing options, but some may be added in a future PR : 

 * DateTimeFormat
 * accessor / mutator prefixes
 * etc ...


src/Attribute/Mapper.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Configures a mapper.
+ */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS)]
+final readonly class Mapper
+{
+    public function __construct(
+        public ?string $source = null,
+        public ?string $target = null,
```
Korbeil, line 17: Shouldn't we change this property names to `acceptedSource` / `acceptedTarget` ? To not be confused with `MapFrom` & `MapTo` properties that are not in the same context.

Korbeil: Also I'm not sure we should accept both properties being set, isn't it a bit confusing ?

tests/Bundle/Resources/App/Api/Entity/Book.php
```diff
@@ -5,12 +5,14 @@
...
 
 use ApiPlatform\Metadata\ApiResource;
+use AutoMapper\Attribute\Mapper;
 use AutoMapper\Attribute\MapTo;
 use AutoMapper\Tests\Bundle\Resources\App\Api\Processor\BookProcessor;
 use AutoMapper\Tests\Bundle\Resources\App\Api\Provider\BookProvider;
 
 /** A book. */
 #[ApiResource(provider: BookProvider::class, processor: BookProcessor::class)]
+#[Mapper(source: 'array', target: 'array')]
```
Korbeil, line 15: So it means we can have a book **from** an array and **to** an array ?

Korbeil (changes requested):
It's also missing some documentation

tests/Bundle/Resources/App/Api/Entity/Book.php
joelwurtz: It means we can map to or from an array

Korbeil (approved):
👍

docs/bundle/cache-warmup.md
```diff
@@ -29,3 +29,38 @@ Then when running the `cache:warmup` command, this will generate the mappers for
...
+```yaml
+automapper:
+  mappings:
+    paths:
+      - "%kernel.project_dir%/src/Entity"
+```
+
+All classes in the specified paths with the `#[Mapper]` attribute will be registered in the container.
+
+This attribute also need a `source` and `target` argument to specify the mapping.
```
Korbeil, line 46: Should be "need a source or/and a target" ?

Korbeil (approved):

PR #117, opened 2024-05-05
feat(attribute): allow multiple source/target, allow overriding attribute with priority system

joelwurtz:
This make those attributes consistent with the way of the `#[Mapper]` attribute works

I think the priority system is better than throwing an exception, and would use case that were not possible before


Korbeil (approved):
👍

PR #124, opened 2024-05-10
Fix WebProfiler issues

### LINKED ISSUE #123: Symfony Web Profiler error: property name on SourcePropertyMetadata
When displaying the Web Profiler panel I get this exception:

> Neither the property "name" nor one of the methods "name()", "getname()"/"isname()"/"hasname()" or "__call()" exist and have public access in class "AutoMapper\Metadata\SourcePropertyMetadata".

If coming from here:

https://github.com/jolicode/automapper/blob/fa61df19e8a05afc9ad28569e4443fdf96034cd4/src/Symfony/Bundle/Resources/views/DataCollector/metadata.html.twig#L123-L128

There is "name" on `property.source` and `property.target`. Replacing "name" by "property" works but I don't know if it's the right fix for this.

## Target

```php
class PostView {
    #[MapFrom(property: 'title')]
    public string $name;
    public string $slug;
    public string $lang;
    public string $content;
}
```

## Source

```php
#[ORM\Entity(repositoryClass: PostRepository::class)]
class Post
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column]
    private ?int $id = null;

    #[ORM\Column(length: 255, unique: true)]
    private ?string $slug = null;

    #[ORM\Column(length: 255)]
    private ?string $title = null;

    #[ORM\Column(type: Types::TEXT)]
    private ?string $content = null;

    #[ORM\Column(length: 6)]
    private ?string $lang = null;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function __construct()
    {
        $this->lang = 'fr';
    }

    public function getSlug(): ?string
    {
        return $this->slug;
    }

    public function setSlug(string $slug): self
    {
        $this->slug = $slug;

        return $this;
    }

    public function getTitle(): ?string
    {
        return $this->title;
    }

    public function setTitle(string $title): self
    {
        $this->title = $title;

        return $this;
    }

    public function getContent(): ?string
    {
        return $this->content;
    }

    public function setContent(string $content): self
    {
        $this->content = $content;

        return $this;
    }

    public function getLang(): ?string
    {
        return $this->lang;
    }

    public function setLang(?string $lang): self
    {
        $this->lang = $lang;

        return $this;
    }
}
```

## Code

```php
// Symfony Bundle service
// AutoMapperInterface $autoMapper

$autoMapper->map($post, PostView::class);
```

Korbeil:
Fixes #123 

We changed some stuff about Metadata which was not impacted within the Symfony WebProfiler

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixes
+- [GH#124](https://github.com/jolicode/automapper/pull/124) Fix Symfony's WebProfiler issues
+
 ## [9.0.0] - 2024-05-06
 ### Added
 - [GH#114](https://github.com/jolicode/automapper/pull/114) Introducing Mapper Attribute
--- a/src/Symfony/Bundle/DataCollector/MetadataCollector.php
+++ b/src/Symfony/Bundle/DataCollector/MetadataCollector.php
@@ -41,7 +41,7 @@ class MetadataCollector extends AbstractDataCollector
             if (class_exists($metadata->mapperMetadata->className)) {
                 $reflectionClass = new \ReflectionClass($metadata->mapperMetadata->className);
 
-                if (($fileName = $reflectionClass->getFileName()) !== false && ($content = file_get_contents($fileName)) !== false) {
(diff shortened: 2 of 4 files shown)


damienalexandre (approved):
Tested locally :+1:

PR #129, opened 2024-05-14
fix(bundle): use map_private_properties when configuring ReflectionExtractor

nikophil:
configuration `map_private_properties` was not used to configure service `automapper.property_info.reflection_extractor` whereas without the bundle, `\AutoMapper\Configuration::$mapPrivateProperties` is actually used, this PR fixes this.

I had to introduce a way to have several configurations available for the fixture bundle, I'm using `KernelTestCase::createKernel()` + an additional config file path for this.

waiting for https://github.com/jolicode/automapper/pull/137

### DIFF
--- a/tests/Bundle/ServiceInstantiationTest.php
+++ b/tests/Bundle/ServiceInstantiationTest.php
@@ -11,6 +11,7 @@ use AutoMapper\Metadata\SourcePropertyMetadata;
 use AutoMapper\Metadata\TargetPropertyMetadata;
 use AutoMapper\Symfony\Bundle\CacheWarmup\CacheWarmer;
 use AutoMapper\Symfony\Bundle\DataCollector\MetadataCollector;
+use AutoMapper\Tests\Bundle\Resources\App\AppKernel;
 use AutoMapper\Tests\Bundle\Resources\App\Entity\AddressDTO;
 use AutoMapper\Tests\Bundle\Resources\App\Entity\ClassWithMapToContextAttribute;
 use AutoMapper\Tests\Bundle\Resources\App\Entity\ClassWithPrivateProperty;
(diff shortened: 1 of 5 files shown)


tests/Bundle/ServiceInstantiationTest.php
```diff
@@ -121,9 +122,10 @@ public function testItCanMapEnums(): void
     /**
      * This test validates that PropertyInfoPass is correctly applied.
      */
-    public function testMapClassWithPrivateProperty(): void
+    public function testMapToClassWithPrivateProperty(): void
     {
-        static::bootKernel();
+        static::bootKernel(['additionalConfigFile' => __DIR__ . '/Resources/config/with-private-properties.yml']);
```
nikophil, line 127: when making `ReflectionExtractor` aware of `map_private_properties`, it has to be set to `true` for this test to pass.

but it seems wrong to me: even if the property is private, it could be written, because it is in the constructor:
```php
class ClassWithPrivateProperty
{
    public function __construct(
        private string $foo
    ) {
    }
}
```

Korbeil (approved):
:+1:

tests/Bundle/ServiceInstantiationTest.php
Korbeil: :warning: We need to fix this behavior before merging this pull request !

joelwurtz: It depends, it only work if we write to it, but not if we read from it

nikophil: > it only work if we write to it, but not if we read from it

it is indeed IMO the expected behavior, even with `map_private_properties: false`, but that's not what happens

joelwurtz: Yes, but not sure we can resolve this easily on our side ?

Does this bug happens because we cannot found the property (not listed) or because we don't find a mutator for it ?

AFAIK, this would require having a different service when we try to get the write mutator on constructor, so we could configure that one to work with private properties

Or do a bug fix in symfony ?

nikophil: ok, forget about that all :)

I've found [a simpler solution](https://github.com/jolicode/automapper/pull/137)

tests/Bundle/ServiceInstantiationTest.php
```diff
@@ -23,15 +24,14 @@
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\KernelInterface;
 
 class ServiceInstantiationTest extends WebTestCase
 {
-    protected function setUp(): void
+    public static function setUpBeforeClass(): void
```
nikophil, line 31: would you mind this change? on my computer it greatly improves test performances:

- before: `Time: 01:53.571`
- after: `Time: 01:11.453`

joelwurtz (approved):

PR #132, opened 2024-05-14
Use use DI Extension class instead of deprecated HttpKernel Extension

GromNaN:
`Symfony\Component\HttpKernel\DependencyInjection\Extension` is deprecated since Symfony 7.1 by https://github.com/symfony/symfony/pull/53801


Korbeil (approved):
Thanks for the contribution @GromNaN ! :ok_hand:

PR #135, opened 2024-05-14
Fix return type of AutoMapper::create()

### LINKED ISSUE #133: Phpstorm seems confused over map function return type
It seems that Phpstorm doesnt understand this line in `AutomapperInterface.php`

`Returns: ($target is class-string|Target ? Target|null : array<mixed>|null)`

I don't know if anything is possible about it or if it's something i did wrong on my side.

See screenshot bellow 

Screenshot from 2024-05-14 14-30-13

lyrixx:
fixed #133


Korbeil (approved):
Thanks ! :pray:

PR #136, opened 2024-05-14
Deprecate the ability to inject AST transformer factories withing stand-alone AutoMapper

Korbeil:


joelwurtz:
Is there really a need for that ?

It requires also removing internal on those interfaces which extends the API a lot for a feature that is really complicated for end user and you can do everything with the current transformer interface ?

Do you have a use case for that ?

Korbeil:
We talked about it with @joelwurtz and decided to remove this feature (deprecating it at first).
This is a complex feature that is really *required* and if it is we could put it into AutoMapper directly.

PR #137, opened 2024-05-15
fix: always allow to write private props in constructor

nikophil:
fixes problem raised [here](https://github.com/jolicode/automapper/pull/129#discussion_r1599499557)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - [GH#158](https://github.com/jolicode/automapper/pull/158) Actually read reload_strategy from bundle configuration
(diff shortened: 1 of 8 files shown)


src/Extractor/SourceTargetMappingExtractor.php
```diff
@@ -17,6 +17,18 @@
  */
 class SourceTargetMappingExtractor extends MappingExtractor
 {
+    public function getProperties(string $class): iterable
+    {
+        return array_values(
+            array_unique(
+                [
+                    ...parent::getProperties($class),
+                    ...$this->getConstructorParameters($class),
```
joelwurtz, line 26: It will fetch properties from the constructor in the source class also which is not wanted in some case

Can we change the signature of this method like : 

```
public getProperties(string $class, bool $withConstructorParamaters = false)
```

Or have two differents methods (one for source, one for target)

And then we would only use it when wanted ? 

(Then we would keep the implementation in the abstract class only)

joelwurtz (approved):

PR #145, opened 2024-05-16
fix(doc): typing error in Configuration for Symfony

TheoD02:
Fixing little mistake in Symfony bundle configuration documentation.

Bundle await for `mapping` but documentation show `mappings`


Korbeil (approved):
Thanks for the typo fix

PR #146, opened 2024-05-16
fix(mapper): Prevent generating GeneratedMapper with special characters

TheoD02:
Hello I've noticed that when we use special characters in `target` name it can prevent them from working.

Here's the error:
image

The code that throw this exception : 
```php
$newObjects = $autoMapper->map($array, ComplexObject::class . '[]');
```

Maybe is not necessary because `map` method don't support `[]` syntax, feel free to close if was a bad idea


Korbeil:
I would prefer something that throw an error when trying to do this over allowing it.

PR #148, opened 2024-05-17
fix: handle deprecated class LNumber in nikic/php-parser v5

nikophil:
I think since the `Automapper` is the only entrypoint of this whole lib, we can add `class_alias()` in this class, and problem would be solved.

WDYT?

### DIFF
--- a/src/php-parser.php
+++ b/src/php-parser.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * This file provides compatibility functions for nikic/php-parser v4 & v5.
+ */
+
+declare(strict_types=1);
+
+namespace AutoMapper\PhpParser;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar;
+
+/**
+ * Constructs an integer number scalar node.
+ *
+ * @param int                  $value      Value of the number
+ * @param array<string, mixed> $attributes Additional attributes
+ */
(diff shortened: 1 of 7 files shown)


joelwurtz:
Not a fan of this, i would keep declaring the alias where it is really needed, not on the top class, main problem is that if we change this later (like some lib that don't depend on the AutoMapper class) we would not have those alias

GromNaN:
Using `class_alias` for classes that doesn't belong to this library can lead to other issues (like an other library checking of the class exist ...). That's bad that php-parser doesn't provide a clean migration path. The best would be that php-parser 4.x would have provided the new classes directly.

A clean alternative would be to create factory functions here, that uses the correct class name depending on the php-parser version.

nikophil:
> A clean alternative would be to create factory functions here, that uses the correct class name depending on the php-parser version.

this was actually what I done in an earlier version :sweat_smile:

joelwurtz (approved):

src/php-parser.php
```diff
@@ -0,0 +1,41 @@
...
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Scalar;
+
+/**
+ * Constructs an integer number scalar node.
+ *
+ * @param int                  $value      Value of the number
+ * @param array<string, mixed> $attributes Additional attributes
```
GromNaN, line 19: This functions should be `@internal`

PR #149, opened 2024-05-17
docs: fix typo

nikophil:
regarding [the declared configuration](https://github.com/jolicode/automapper/blob/main/src/Symfony/Bundle/DependencyInjection/Configuration.php#L56) the right key is `mapping` and not `mappings`


Korbeil (approved):
Crap https://github.com/jolicode/automapper/pull/145 didn't fixed all of them, thanks :pray:

PR #158, opened 2024-05-29
fix(bundle): actually read reload_strategy from bundle configuration

nikophil:
Currently, because of parameter precedence, `$reloadStrategy` is never read from bundle config and always default to `ALWAYS` :see_no_evil:


Korbeil (approved):
Damn, thanks for the fix, could you add a note about it in the CHANGELOG pls ? :pray:

PR #166, opened 2024-06-20
feat(symfony): remove useless pass for property info, remove cache

joelwurtz:
I remove the compiler pass for property info, i think it was useless.

I also removed the cache for property info, we don't need it and it was causing DX issues when changing metadata of a class (as data was cached and not correctly reload)

Last thing i used only our own services for property info, i don't want that other configuration messing up with our way of configuring those classes


Korbeil (approved):
A quick CHANGELOG note ? :pray:

PR #174, opened 2024-07-16
feat(loader): lock load class on file loader to avoid race condition

joelwurtz:
There can be race condition when multiple php process would generate the class at the same time, so we add a lock here to avoid this problem

[Better diff with `?w=1`](https://github.com/jolicode/automapper/pull/174/files?w=1)

### DIFF
--- a/src/Loader/FileLoader.php
+++ b/src/Loader/FileLoader.php
@@ -11,13 +11,14 @@ use AutoMapper\Metadata\MetadataFactory;
 use AutoMapper\Metadata\MetadataRegistry;
 use PhpParser\PrettyPrinter\Standard;
 use PhpParser\PrettyPrinterAbstract;
+use Symfony\Component\Lock\LockFactory;
 
 /**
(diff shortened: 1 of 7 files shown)


lyrixx (approved):
Awesome, thanks

src/Loader/FileLoader.php
```diff
@@ -40,25 +42,33 @@ public function loadClass(MapperMetadata $mapperMetadata): void
         $className = $mapperMetadata->className;
         $classPath = $this->directory . \DIRECTORY_SEPARATOR . $className . '.php';
 
-        if ($this->reloadStrategy === FileReloadStrategy::NEVER && file_exists($classPath)) {
-            require $classPath;
+        // We lock the file here, because another process could be writing the file at the same time
```
Korbeil, line 45: Shouldn't we wait until the _other_ process is done until returning this method ?
So we are sure the file is generated.

PR #179, opened 2024-08-14
fix(doc): Updating getting-started default constructor strategy

Ngob:
Hello,

I noticed that you changed the default constructor strategy from "always" to "auto", so I updated the documentation

I hope that's OK for you

Have a great day


Korbeil:
Thanks for the typo :)

PR #180, opened 2024-08-30
feat: option to generate code with strict types

MrMeshok:
PHP, by default, will coerce values of the wrong type into the expected scalar type declaration if possible. This leads to unexpected behavior. For example, losing decimal digits when casting float to int
```PHP
readonly class IntDto
{
    public function __construct(
        public int $var,
    ) {}
}

$autoMapper = \AutoMapper\AutoMapper::create();
$result = $autoMapper->map(['var' => 1.1111], IntDto::class);
$result->var; // var is 1
```
This PR adds configuration to add `declare(strict_types=1)` to generated code when mapping, so instead of unexpected behavior, you get TypeError.

If everything is ok, I will update the docs, but I am not very fluent with english, so maybe someone else could write a more understandable description.

### DIFF
--- a/src/Metadata/GeneratorMetadata.php
+++ b/src/Metadata/GeneratorMetadata.php
@@ -24,7 +24,8 @@ final class GeneratorMetadata
         public readonly array $propertiesMetadata,
         public readonly bool $checkAttributes = true,
         public readonly ConstructorStrategy $constructorStrategy = ConstructorStrategy::AUTO,
-        public bool $allowReadOnlyTargetToPopulate = false,
+        public readonly bool $allowReadOnlyTargetToPopulate = false,
+        public readonly bool $strictTypes = false,
         public readonly ?string $provider = null,
     ) {
         $this->variableRegistry = new VariableRegistry();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- [GH#180](https://github.com/jolicode/automapper/pull/180) Add configuration to generate code with strict types
+
 ### Fixed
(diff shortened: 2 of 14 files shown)


joelwurtz:
:+1: for this thanks, can you add the possibility to configure this through the `Mapper` attribute also ?

Also wondering if declare strict should be true by default ? poke @Korbeil @nikophil ?

If you could update the doc and add a note in the changelog it would be nice, if you are not comfortable doing so we can do it.

Korbeil:
I'm totally for this option and I think we should put it by default as @joelwurtz stated.
Same as he said, we would require a configuration on the `Mapper` attribute :pray:

nikophil:
won't we face some BC break if this becomes a default thing?

MrMeshok:
Made it configurable from `Mapper` and enabled by default. I'll wait for the decision about the default value and force push if it needs to be changed. I will leave the changelog and docs for you

src/Metadata/GeneratorMetadata.php
```diff
@@ -24,7 +24,8 @@ public function __construct(
         public readonly array $propertiesMetadata,
         public readonly bool $checkAttributes = true,
         public readonly ConstructorStrategy $constructorStrategy = ConstructorStrategy::AUTO,
-        public bool $allowReadOnlyTargetToPopulate = false,
+        public readonly bool $allowReadOnlyTargetToPopulate = false,
```
MrMeshok, line 27: Seems like readonly was missed here

Korbeil:
> won't we face some BC break if this becomes a default thing?

This is right, we should put it disabled by default and we will put it enabled by default on next major ~
Could you make the change @MrMeshok ? :pray:

MrMeshok:
@Korbeil Done👍

Korbeil (approved):
Looks good to me :ok_hand:

Could you add a note about your feature into the CHANGELOG please ?

MrMeshok:
Added note to CHANGELOG and combined everything in 1 commit

Korbeil:
Thanks for the feature @MrMeshok ! :pray:

PR #183, opened 2024-09-04
feat: Ability to change reload strategy from AutoMapper::create()

MrMeshok:


joelwurtz:
Could this use the `Configuration` class instead of being a parameter, i think we should go this way for every configuration parameter, it will be easier for BC break rather than having a new parameter

However you may need to change the symfony bundle in order to make it work (not sure)

MrMeshok:
Yea, I thought about putting it in `Configuration`, but create method already has cacheDirectory argument, so it seems natural to put reloadStrategy there too. Also, I did have a look in the symfony bundle, as far as I can tell, `AutoMapper` creates from DI without using create method, and reloadStrategy from `FileLoader` gets replaced through container

Korbeil (changes requested):
Thanks for this feature @MrMeshok, could you move the parameter to the `Configuration` class and add a note about your change in the CHANGELOG please ? :pray:

MrMeshok:
@Korbeil Done. I also did a small change in Symfony bundle, although it is not necessary

Korbeil (approved):

Korbeil:
Thanks for your contribution @MrMeshok !

PR #184, opened 2024-09-12
fix: error when mapping from stdClass to constructor with nullable/optional arguments

### LINKED ISSUE #181: Undefined property error when mapping from stdClass to class with nullable properties
```PHP
readonly class Response
{
    public function __construct(
        public int $ResponseCode,
        public ?string $ResponseDescription = null,
    ) {}
}

$data = new \stdClass();
$data->ResponseCode = 0;

$dto = $autoMapper->map($data, Response::class);
```
With this code, mapper will generate code like this:
```PHP
$value_1 = null;
if (null !== $value->ResponseDescription) {
    $value_1 = $value->ResponseDescription;
}
$constructarg_1 = $value_1 ?? NULL;
```
And we will get an undefined property error because ResponseDescription is not guaranteed to exist in the source.
But if data is an array, the generated code looks like this:
```PHP
$value_1 = null;
if (isset($value['ResponseDescription'])) {
    $value_1 = $value['ResponseDescription'];
}
$constructarg_1 = $value_1 ?? NULL;
```
So when source is an object and target property is nullable or has default value we need to do this:
```PHP
$constructarg_1 = $value->ResponseDescription ?? NULL;
```
Also, array code can be rewritten too
```PHP
$constructarg_1 = $value['ResponseDescription'] ?? NULL;
```
And I can work on implementation myself, I already did some testing

MrMeshok:
I was trying to understand why generated code looks weird when mapping from array/stdClass to class which has constructor with nullable/optional arguments.
NullableTransformer is not supposed to be applied when creating from constructor because nullable/optional already handled.

With this fix, generated code instead of looking like this:
```PHP
$value_1 = null;
if (null !== $value->ResponseDescription) {
    $value_1 = $value->ResponseDescription;
}
$constructarg_1 = $value_1 ?? NULL;
``` 
Look like this:
```PHP
$constructarg_1 = $value->ResponseDescription ?? NULL;
```

Which also fixes #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - [GH#180](https://github.com/jolicode/automapper/pull/180) Add configuration to generate code with strict types
 
+### Fixed
+- [GH#184](https://github.com/jolicode/automapper/pull/184) Fix error when mapping from stdClass to constructor with nullable/optional arguments
+
 ## [9.1.2] - 2024-09-03
 ### Fixed
 - [GH#174](https://github.com/jolicode/automapper/pull/174) Fix race condition when writing generated mappers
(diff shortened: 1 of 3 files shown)


Korbeil (approved):
Thanks @MrMeshok for this fix, it looks nice ! :heavy_check_mark: Could you also add a note in the CHANGELOG about your change please ? :pray:

MrMeshok:
@Korbeil Done. Can you merge it today? I am working on another PR and don't want to deal with conflicts in CHANGELOG😁

Korbeil:
@MrMeshok merged, thanks for your contribution ! 🙏

MrMeshok:
@Korbeil I got regression errors after this fix with classes that have nullable objects as arguments. I'm working on fix and tests for this

PR #187, opened 2024-09-13
fix: regression after #184

MrMeshok:
Turns out NullableTransformer is needed when creating from constructor if we have argument what expects nullable object


Korbeil:
Quick CHANGELOG line ? :)

MrMeshok:
Something like that?
[GH#187](https://github.com/jolicode/automapper/pull/187) Fix regression after [GH#187](https://github.com/jolicode/automapper/pull/187)

Korbeil:
Yeah looks fine :+1:

Korbeil (approved):

PR #188, opened 2024-09-13
fix: constructor with default values as objects

MrMeshok:
While working on #187 I noticed what you can't map to a class that has constructor with default values as objects if source missing these arguments.
Seems like a complex issue to fix, so I'll start with the tests

### DIFF
--- a/src/Generator/CreateTargetStatementsGenerator.php
+++ b/src/Generator/CreateTargetStatementsGenerator.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace AutoMapper\Generator;
 
-use AutoMapper\Exception\CompileException;
 use AutoMapper\Exception\MissingConstructorArgumentsException;
 use AutoMapper\Generator\Shared\CachedReflectionStatementsGenerator;
 use AutoMapper\Generator\Shared\DiscriminatorStatementsGenerator;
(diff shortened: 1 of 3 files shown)


MrMeshok:
Actually, there is a simple way. 
Instead of this:
```php
$constructorarg_1 = $source->author ?? Author::__set_state();

$item = new Post(author: $constructorarg_1)
```
It should be this:
```php
$constructorargs = [];
if (isset($source->author)) {
    $constructorargs['author'] = $source->author;
}

$item = new Post(...$constructorargs)
```

MrMeshok:
@Korbeil Sorry for unrelated question, I am just not sure how to reach out.
I was wondering when is the next release? I need some fixes from main, but don't really want to lower minimum stability in composer and use commit/branch

joelwurtz:
Will look at it, i think indeed using named parameters would avoid a lot of problems

MrMeshok:
@joelwurtz I did some testing, works great, but behavior for checking if value exists changed.
Before it was using null coalescing operator aka isset(), now it's array_key_exists/property_exists. And I was kinda depending on isset checks. Sometimes you work with APIs what have parameters which can be unreasonably nullable. But in your DTO you can just write it as property with default value and skip null checks in your code. 
```
class SimpleDto
{
    public function __construct(
        public int $stringProperty = '',
    ) {}
}

$source = ['stringProperty' => null];

$autoMapper->map($source, SimpleDto::class);
```
Now this will fail. And also isset would be faster

joelwurtz:
I have add https://github.com/jolicode/automapper/pull/236 to fix that however it does not use isset everywhere, as there was a bug on that before so know : 

 * use isset if construct arg does not allow nullable value
 * use array_key_exists / property_exists if construct arg allow nullable value

PR #192, opened 2024-10-03
fix: source and context not passed to callable transformer

### LINKED ISSUE #191: Source and context parameters not passed to transformer ?
Hi all,

following the documentation I tried a static callback transformer. This is the simple code :

```php
<?php

require __DIR__.'/vendor/autoload.php';

use AutoMapper\Attribute\MapFrom;
use AutoMapper\AutoMapper;

class Source
{
    public ?int $id = null;
}

class Target
{
    #[MapFrom(source: Source::class, transformer: [self::class, 'transformId'])]
    public ?string $id = null;

    public static function transformId(int $id, Source $source, array $context): string
    {
        return 'https://example.com/' . $id;
    }
}
$automapper = AutoMapper::create();

$source = new Source();
$source->id = 42;

$target = $automapper->map($source, new Target());
```

But I have the following error

```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function Target::transformId(), 1 passed in [...]\vendor\jolicode\automapper\src\Loader\EvalLoader.php(36) : eval()'d code on line 19 and exactly 3 expected in [...]\index.php on line 18
```
It seems like Source and context are not sent to the transformer like the doc says.

Did I miss something ?

Thank you

MrMeshok:
Fixes #191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [GH#184](https://github.com/jolicode/automapper/pull/184) Fix error when mapping from stdClass to constructor with nullable/optional arguments
 - [GH#185](https://github.com/jolicode/automapper/pull/185) Fix constructor with default parameter array does not work with constructor_arguments context
 - [GH#187](https://github.com/jolicode/automapper/pull/187) Fix regression after [GH#184](https://github.com/jolicode/automapper/pull/184)
+- [GH#192](https://github.com/jolicode/automapper/pull/192) Fix source and context not passed to callable transformer
 
 ## [9.1.2] - 2024-09-03
 ### Fixed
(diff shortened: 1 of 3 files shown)


Korbeil (approved):
Looks good, could you add a test to validate the test ? :pray:

MrMeshok:
It's already should be tested, because class Bar is used in generic MapTo/MapFrom tests

Korbeil:
I mean, it should fail if it is tested, you didn't change a test in this pull request and clearly it didn't failed before.

MrMeshok:
I changed `tests/Fixtures/MapTo/Bar.php` now `transformC` and `transformDStatic` requires source and context. Also I added `transformer: 'htmlspecialchars'` to $form property so usage of internal functions tested too. Without fix in `CallableTransformer` test will fail with this additions to `Bar`

Korbeil (approved):
Thanks for the test and your contribution! @MrMeshok 🙏

PR #194, opened 2024-10-07
Refacto: Remove the lib phpdocumentor/type-resolver that is not used

p-pichet:
When we use the automapper on a Symfony project using ApiPlatform, we start having issue with the phpdocumentor. 

Base on my research, it's not use on the project and should be comletely remove.


p-pichet:
Sorry. 
I close this PR as it's not the automapper issue

PR #197, opened 2024-10-11
Fix race condition in EvalLoader

MrMeshok:
When using swoole coroutines with eval loader, where is a possibility to get an error `Cannot declare class {class}, because the name is already in use`, because coroutines share memory and declared classes.
Can be tested like this
```php
require 'vendor/autoload.php';

readonly class Dto
{
    public function __construct(
        public int $id,
    ) {}
}

$mapper = \AutoMapper\AutoMapper::create();
$data = ['id' => 1];

\Co\run(static function () use ($mapper, $data) {
    for ($i=1; $i < 100_000; $i++) { 
        \Co\go(static fn () => $mapper->map($data, Dto::class));
    }
});
```

### DIFF
--- a/src/Loader/EvalLoader.php
+++ b/src/Loader/EvalLoader.php
@@ -16,24 +16,39 @@ use PhpParser\PrettyPrinterAbstract;
  *
  * @author Joel Wurtz <jwurtz@jolicode.com>
  *
- *  @internal
+ * @internal
  */
-final readonly class EvalLoader implements ClassLoaderInterface
+final class EvalLoader implements ClassLoaderInterface
 {
-    private PrettyPrinterAbstract $printer;
+    private static array $lockMap = [];
 
     public function __construct(
-        private MapperGenerator $generator,
-        private MetadataFactory $metadataFactory,
+        private readonly MapperGenerator $generator,
+        private readonly MetadataFactory $metadataFactory,
+        private readonly PrettyPrinterAbstract $printer = new Standard(),
     ) {
-        $this->printer = new Standard();
     }
 
     public function loadClass(MapperMetadata $mapperMetadata): void
     {
-        eval($this->printer->prettyPrint($this->generator->generate(
-            $this->metadataFactory->getGeneratorMetadata($mapperMetadata->source, $mapperMetadata->target)
-        )));
+        if (isset(self::$lockMap[$mapperMetadata->className])) {
+            do {
+                usleep(100000); // 0.1 second
+            } while (isset(self::$lockMap[$mapperMetadata->className]));
+
+            if (class_exists($mapperMetadata->className, false)) {
+                return;
+            }
+        }
+
+        self::$lockMap[$mapperMetadata->className] = true;
+        try {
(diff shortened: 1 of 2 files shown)


src/Loader/EvalLoader.php
```diff
@@ -16,24 +16,40 @@
...
-        )));
+        if (isset(self::$lockMap[$mapperMetadata->className])) {
+            do {
+                usleep(100000); // 0.1 second
+            } while (isset(self::$lockMap[$mapperMetadata->className])); // @phpstan-ignore isset.offset
+
+            if (class_exists($mapperMetadata->className, false)) {
+                return;
+            }
+        }
```
MrMeshok, line 43: This code block looks a little weird, but there is a reason for this.
At first, it looked like this
```php
while (isset(self::$lockMap[$mapperMetadata->className])) {
    usleep(100000); // 0.1 second
}

if (class_exists($mapperMetadata->className, false)) {
    return;
}
```
But there is a problem, class_exists gets called every time, but he only needed if lock existed, so I restructured code to first check lock inside if and then start while loop

joelwurtz:
Indeed, we do have done something similar for the FileLoader here https://github.com/jolicode/automapper/pull/174

Can you use the LockFactory instead of a while loop, like done on this PR ? This way i believe you could use a specific implementation that can handle an event loop for the LockFactory (like one that use this https://www.php.net/manual/fr/class.swoole-lock.php)

Then it will avoid loop + sleep and just block in an async way (allowing you to execute other coroutines)

src/Loader/EvalLoader.php
joelwurtz: Not sure you could do that if using the LockFactory, anyway an extra call to class_exists should not be a problem

MrMeshok:
I did some digging, turns out using Swoole\Lock inside coroutines is a [bad idea](https://github.com/deminy/swoole-by-examples/blob/c80719c1e8f916bf71e652b962007ac7213b5227/examples/csp/deadlocks/swoole-lock.php), instead you should use [channel](https://github.com/deminy/swoole-by-examples/blob/c80719c1e8f916bf71e652b962007ac7213b5227/examples/io/block-a-coroutine.php). I also had problems with FileLoader, it would cause deadlocks. I did some research, coroutines work fine with flock(), it's just specific implementation of FlockStore not compatible with coroutines. But am not gonna resume this PR, as it feels weird to put swoole dependent code in general PHP package.

Instead I created CoroutineAwareLoader which encapsulates FileLoaded\EvalLoader and if it's used in coroutine context locks other coroutines through channels. To do this I had to do Closure::bind() on AutoMapper, to access private properties, which is not ideal, but it works.

Also after upgrading to Swoole 6 I could'nt reproduce this with simple run + go (deadlocks in FileLoader are reproducible), so I had to setup simple Swoole server and send parallel request to it. This ads some IO and context switching to coroutines and reproduces the error (but not on every try). And it seems with PHP 8.4 error message for declaring classes which already exists changed a little.

PR #199, opened 2024-10-16
fix: signature of Map attributes `transformer` property

JorickPepin:
According to the `PropertyTransformerInterface`, the signature of a transformer should be: 

```php
/**
 * @param mixed                       $value   the value of the property to transform, can be null if there is no way to read the data from the mapping
 * @param object|array<string, mixed> $source  the source input on which the custom transformation applies
 * @param array<string, mixed>        $context Context during mapping
 */
public function transform(mixed $value, object|array $source, array $context): mixed;
```

### DIFF
--- a/src/Attribute/MapFrom.php
+++ b/src/Attribute/MapFrom.php
@@ -11,14 +11,14 @@ namespace AutoMapper\Attribute;
 final readonly class MapFrom
 {
     /**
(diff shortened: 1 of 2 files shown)


joelwurtz (approved):

joelwurtz:
Thanks @JorickPepin

PR #200, opened 2024-10-22
Added skip unitialized values context param

Korbeil:


TheoD02:
Hello, sorry for the late testing phase 😁

PR need a rebase to fix:

```
Too few arguments to function PHPStan\\PhpDocParser\\Parser\\ConstExprParser::__construct(), 0 passed in /app/vendor/jolicode/automapper/src/Extractor/GetTypeTrait.php on line 46 and exactly 1 expected
```

But apart of this error, everything work like a charm, thanks !

Korbeil:
I rebased the PR @TheoD02.

But I'm kinda blocked with this pull request, in one of the test (see `AutoMapper::testSkipNullValues`) we have a class property default value to `null`. So even with `isset` we cannot detect if the property was initialized or not since it will return `false` because it's `null` :disappointed:

TheoD02:
> I rebased the PR @TheoD02.
> 
> But I'm kinda blocked with this pull request, in one of the test (see `AutoMapper::testSkipNullValues`) we have a class property default value to `null`. So even with `isset` we cannot detect if the property was initialized or not since it will return `false` because it's `null` 😞
> 
> Here is the issue: https://3v4l.org/BsDTZ

Hello thanks for your answer, maybe we can do something like in PropertyAccessor ?

Not really checked the possibility in your implementation

https://github.com/symfony/property-access/blob/b28732e315d81fbec787f838034de7d6c9b2b902/PropertyAccessor.php#L417-L429

Korbeil:
After talking with @lyrixx he gave me the same thing that PropertyAccess does use: https://3v4l.org/idiOR
I'll try using that trick.

Korbeil:
@TheoD02 the issue is now fixed, if you want to test again / review that PR !

TheoD02:
> @TheoD02 the issue is now fixed, if you want to test again / review that PR !

I'll will check this evening, or tomorrow 😀

Korbeil:
Any news @TheoD02 ?

TheoD02:
> Any news @TheoD02 ?

Sorry, didn't have time to test it :/

Just updated and runned some tests, and no regression identified on my project, seem to be good ! 😄

PR #205, opened 2024-11-15
Add support for phpstan/phpdoc-parser 2

MrMeshok:
I noticed locally that random test started to fail, turns out it's because [phpdocumentor/type-resolver added support for phpstan/phpdoc-parser v2](https://github.com/phpDocumentor/TypeResolver/releases/tag/1.10.0), but this library not ready for phpdoc-parser v2.
I chose minimal version of 1.13 because minimal version of [phpdocumentor/type-resolver required it](https://github.com/phpDocumentor/TypeResolver/blob/1.7.0/composer.json#L15) so there should be no changes to `composer update --prefer-lowest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - [GH#186](https://github.com/jolicode/automapper/pull/186) Optimize creation from constructor
+- [GH#205](https://github.com/jolicode/automapper/pull/205) Add support for phpstan/phpdoc-parser 2
 
 ### Fixed
 - [GH#184](https://github.com/jolicode/automapper/pull/184) Fix error when mapping from stdClass to constructor with nullable/optional arguments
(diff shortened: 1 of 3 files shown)


Korbeil:
Is it not possible to have something compatible like Symfony did with TypeInfo with https://github.com/symfony/symfony/pull/58804 ?

MrMeshok:
> Is it not possible to have something compatible like Symfony did with TypeInfo with [symfony/symfony#58804](https://github.com/symfony/symfony/pull/58804) ?

It's definitely possible, I just did quick fix, this issue probably gonna show up then someone does `composer update` in their project. Also phpstan/phpdoc-parser should stay in composer.json so this doesn't happen again

MrMeshok:
@Korbeil I changed PR to add support to v2 instead of locking to v1

Korbeil (approved):
Looks perfect, thanks for the update! I'll try to get a look today to do a release!

MrMeshok:
> Now, do I read this issue correctly in that support for 2.x is added but has yet to be released?

Yes.
You can add `"phpstan/phpdoc-parser": "^1.13"` to your composer.json as workaround for now

Korbeil:
I just released AutoMapper `9.2.0` which should resolve your issue without forcing v1 on `phpstan/phpdoc-parser` @jorismak https://github.com/jolicode/automapper/releases/tag/9.2.0

PR #207, opened 2024-11-25
fix: implicity nullable parameter deprecations

JorickPepin:
Deprecated since PHP 8.4, see https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


Korbeil:
Thanks for the changes, could you add a quick note about it in the CHANGELOG pls ?

Korbeil:
Thanks for the update @JorickPepin !

bpolaszek:
Hi @Korbeil, would you mind tagging `9.2.1` with that one? 🙃

Issue #211, opened 2024-12-06
Cannot map to a class that has adder and remover AND constructor arguments


wuchen90:
When the target class has adder and remover for a property and if we have this said property as constructor arguments also, the mapper will break because the `\AutoMapper\Transformer\AbstractArrayTransformer` will check for adder and try to use it.

In case where the target is just a classname and not an instance, an error will be thrown.

```
Undefined variable $constructarg
```

I tried to fix it by my own but my knowledge is limited on this matter, so I just made a test to reproduce this issue: https://github.com/wuchen90/automapper/pull/1

PR #212, opened 2024-12-06
fix: handle case where class target has adder and remover AND constructor arguments

wuchen90:
After some attempts, I came to this solution for this issue: https://github.com/jolicode/automapper/issues/211.

Not sure if it's the right thing to do.


Korbeil (approved):
Looks good to me, could you write a quick thing in CHANGELOG about your changes please 🙏

Korbeil:
Thanks for your contribution @wuchen90 !

PR #225, opened 2025-02-17
Add mapCollection method to base interface

Korbeil:
I had a lot of feedback that people wants to map a collection of items into something else and most of the time they don't really like the fact they have to do the `foreach` themselves.

So here is a `mapCollection(array $collection, string $target, array $context = []): array` method that allows mapping collection of items.


src/AutoMapperInterface.php
```diff
@@ -24,4 +24,16 @@ interface AutoMapperInterface
...
+     * @template Source of object
+     * @template Target of object
+     *
+     * @param array<array<mixed>>|array<Source> $collection
+     * @param class-string<Target>|'array'      $target
+     * @param MapperContextArray                $context
+     *
+     * @return ($target is class-string ? array<Target> : array<mixed>)
+     */
+    public function mapCollection(array $collection, string $target, array $context = []): array;
```
nikophil, line 38: maybe we could map any iterable?
```suggestion
    public function mapCollection(iterable $collection, string $target, array $context = []): array;
```

nikophil (approved):

PR #230, opened 2025-03-04
Fix nested property when it's an array without type indicator on only one of the source or target

lyrixx:


tests/Fixtures/IssueGreg/ComponentDto.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AutoMapper\Tests\Fixtures\IssueGreg;
```
lyrixx, line 5:
```suggestion
namespace AutoMapper\Tests\Fixtures\Issue230;
```

and so on?

lyrixx (commented):
Thank you very much for the quick fix!

PR #231, opened 2025-03-04
fix(constructor): better check for property exists when building constructor args

lyrixx:


tests/AutoMapperTest.php
```diff
@@ -475,6 +475,43 @@ public function testConstructor(): void
         self::assertTrue($userDto->getConstructor());
     }
 
+    public function testConstructorAndRelationMissing(): void
+    {
+        $user = ['name' => 'foo'];
+        $this->expectException(MissingConstructorArgumentsException::class);
```
joelwurtz, line 481: i added this, from the input this is the expected use case

tests/AutoMapperTest.php
```diff
@@ -475,6 +475,43 @@ public function testConstructor(): void
...
+
+        /** @var Fixtures\UserConstructorDTOWithRelation $userDto */
+        $userDto = $this->autoMapper->map($user, Fixtures\UserConstructorDTOWithRelation::class);
+    }
+
+    public function testConstructorAndRelationMissing2(): void
+    {
+        $user = ['name' => 'foo', 'int' => ['foo' => 1]];
```
joelwurtz, line 489: I changed the input array, passing an Object will not work as we expecte pure base type when mapping from array, if there is something else then a transformer should be provided

tests/AutoMapperTest.php
```diff
@@ -475,6 +475,43 @@ public function testConstructor(): void
...
+        self::assertSame(30, $userDto->age);
+    }
+
+    public function testConstructorAndRelationMissingAndContext(): void
+    {
+        $user = ['name' => 'foo'];
+        /** @var Fixtures\UserConstructorDTOWithRelation $userDto */
+        $userDto = $this->autoMapper->map($user, Fixtures\UserConstructorDTOWithRelation::class, [
+            MapperContext::CONSTRUCTOR_ARGUMENTS => [
+                Fixtures\UserConstructorDTOWithRelation::class => ['int' => new Fixtures\IntDTO(1)],
```
joelwurtz, line 505: class was wrong, i put the correct one and it works as expected

lyrixx:
Thanks 👍🏼

PR #232, opened 2025-03-05
chore(tools): use castor for cs fixer and phpstan instead of tools directory

joelwurtz:
Still in draft, need to have a new release of castor with somes fixs to works


joelwurtz:
ping @nikophil @Korbeil this ease running our tools and should produce less maintenance

Korbeil (commented):
With this, how do we update PHP-CS-Fixer or PHPStan ? It is managed by `castor-php/php-qa` package ?

joelwurtz:
You just have to update the version defined here https://github.com/jolicode/automapper/pull/232/files#diff-4914903c5909600433aedbfbbb5204713647759c3232f6d09002729e65925ab0R13

Korbeil (approved):

PR #240, opened 2025-03-10
Add failing test case for constructor + readonly + list + constructor_arguments

lyrixx:


joelwurtz:
It's an intended behavior, see https://automapper.jolicode.com/latest/getting-started/configuration/ on `constructStrategy` 

Adding 

```php
#[Mapper(constructorStrategy: ConstructorStrategy::ALWAYS)]
```

on top of class make the test work

By default, if all properties from the source are not present on the target constructor we build the class without invoking the constructor

Otherwise we generate it, this parameter force the generation of the constructor.

Maybe we could always generate it and use this behavior depending on a context value instead ?

lyrixx:
It's up to you, but what is sure is I don't need this feature

joelwurtz:
> It's up to you, but what is sure is I don't need this feature

Could be useful if there is code that can give those constructor arguments and other that cannot for the same mapping

joelwurtz:
I'm closing this, don't have the time to do that and even if the feature can be useful, i never needed it.

PR #252, opened 2025-04-01
Add support for SerializedName attributes

lyrixx:
I, This is just the test for now, and it fails.

If someone could finish this PR, It would be awesome 😍


lyrixx:
Thank tout very much !

PR #260, opened 2025-04-04
feat(mapper): allow to merge existing values by extracting identifiers

joelwurtz:
This introduce the possibility to deep merge object with collections and fetching the correct value to update it instead of adding a new one into the collection

### DIFF
--- a/src/Extractor/WriteMutator.php
+++ b/src/Extractor/WriteMutator.php
@@ -32,8 +32,8 @@ final class WriteMutator
 
     public function __construct(
         public readonly int $type,
-        private readonly string $property,
-        private readonly bool $private = false,
+        public readonly string $property,
+        public readonly bool $private = false,
         public readonly ?\ReflectionParameter $parameter = null,
     ) {
     }
--- a/src/Transformer/AbstractArrayTransformer.php
+++ b/src/Transformer/AbstractArrayTransformer.php
@@ -27,7 +27,7 @@ abstract readonly class AbstractArrayTransformer implements TransformerInterface
 
     abstract protected function getAssignExpr(Expr $valuesVar, Expr $outputVar, Expr $loopKeyVar, bool $assignByRef): Expr;
 
-    public function transform(Expr $input, Expr $target, PropertyMetadata $propertyMapping, UniqueVariableScope $uniqueVariableScope, Expr\Variable $source): array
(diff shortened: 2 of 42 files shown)


src/Extractor/WriteMutator.php
```diff
@@ -32,8 +32,8 @@ final class WriteMutator
 
     public function __construct(
         public readonly int $type,
-        private readonly string $property,
-        private readonly bool $private = false,
+        public readonly string $property,
+        public readonly bool $private = false,
         public readonly ?\ReflectionParameter $parameter = null,
         private readonly ?string $removeMethodName = null,
```
Korbeil, line 38: Can't we pass this as public as well ? Since it's readonly there is no point letting it alone as private.

src/Transformer/AbstractArrayTransformer.php
```diff
@@ -68,28 +62,70 @@ public function transform(Expr $input, Expr $target, PropertyMetadata $propertyM
...
+                $itemStatements[] = new Stmt\Expression(new Expr\Assign($existingValue, new Expr\BinaryOp\Coalesce(new Expr\ArrayDimFetch($exisingValuesIndexed, $hashValueTargetVariable), new Expr\ConstFetch(new Name('null')))));
+            }
             $itemStatements[] = new Stmt\Expression(new Expr\Assign($mappedValueVar, $output));
             $itemStatements[] = new Stmt\If_(new Expr\BinaryOp\NotIdentical(new Expr\ConstFetch(new Name('null')), $mappedValueVar), [
                 'stmts' => [
                     new Stmt\Expression($propertyMapping->target->writeMutator->getExpression($target, $mappedValueVar, $assignByRef)),
                 ],
             ]);
+
+        // @TODO handle existingValue
```
Korbeil, line 100: ?

Korbeil (commented):
I do think it adds a lot of complexity to Array transformers (from/to), would it be possible to add some kind of abstraction to separate the pure mapping part & the identifier/existing values handling in separate class / traits or whatever ?

joelwurtz:
> I do think it adds a lot of complexity to Array transformers (from/to), would it be possible to add some kind of abstraction to separate the pure mapping part & the identifier/existing values handling in separate class / traits or whatever ?

Yeah but all of this is internal and it's only on implementation, i don't want to over abstract thing, as it may changes and other bugs may appear, we can always change it latter since it's internal, i first want feedback on using this feature before abstracting the code and public api is minimal on this (so it would not impact anything)

Issue #227, opened 2025-02-24
Mapping arrays to arrays differs if using direct assignment vs adder/remover


allflame:
There is incosistency in behavior between
```
class From
{
    public array $values;
}

class To
{
    public array $values;
}

$from = new From();
$to = new To();
$from->values = [1, 2, 3];
$to->values = [4, 5, 6];
$result = AutoMapper::create()->map($from, $to);
// $result->values is [1, 2, 3]
```
and approach with private/protected property and adder/remover
```
class From
{
    public array $values;
}

class To
{
    private array $values;

    public function addValue(mixed $value): void
    {
        $this->values[] = $value;
    }

    public function removeValue(mixed $value): void
    {
        foreach ($this->values as $key => $existing) {
            if ($value === $value) {
                unset($this->values[$key]);
            }
        }
    }

    public function getValues(): array
    {
        return $this->values;
    }
}

$from = new From();
$to = new To();
$from->values = [1, 2, 3];
$to->values = [4, 5, 6];
$result = AutoMapper::create()->map($from, $to);
// $result->values is [4, 5, 6, 1, 2, 3]
```
I believe this is also the cause for the issue outlined [here](https://github.com/jolicode/automapper/issues/169#issuecomment-2599064980)

It seems like removerInfo is discarded and never used in the [MappingExtractor](https://github.com/jolicode/automapper/blob/main/src/Extractor/MappingExtractor.php#L130) leading to duplications.
Real use-case problem is mapping relations x-to-Many relations from DTOs to Doctrine collections, whereby whatever the expected behavior might be (and problems associated with it [169](https://github.com/jolicode/automapper/issues/169), [198](https://github.com/jolicode/automapper/issues/198), ), it is definitely not a duplication.

allflame:
Hey @joelwurtz . Thanks for taking the time to look into this issue. If you ask me, I'd say the expected behavior is to have both array to be equal (in a sense of sets, not identical). So arrays `[1, 2]` and `[2, 1]` would be considered the same. Main reason is that I think to achieve "identity" in case of accessors (through adder/remover) might be tricky and inefficient. Having this in mind let me try to answer your questions

- Mapper tries to map from source to target, and not enrich the target with the source, so any two values, irrespective of type will be considered equivalent. So I firmly believe current behavior in case of direct array to array mapping is the expected one.
- Not sure what you meant by undefined property. If the target doesn't have a respective property? In this case I though the mapper is just skipping it? I don't see how arrays are special here.
[...]

PR #261, opened 2025-04-04
feat(remover): remove existing values when adder and remover

joelwurtz:
Fix https://github.com/jolicode/automapper/issues/227

### DIFF
--- a/tests/AutoMapperTest.php
+++ b/tests/AutoMapperTest.php
@@ -839,13 +839,11 @@ class AutoMapperTest extends AutoMapperTestCase
         $this->autoMapper->map($petOwnerAsArray, $petOwner);
 
         self::assertIsArray($petOwner->getPets());
-        self::assertCount(3, $petOwner->getPets());
-        self::assertSame('Nemo', $petOwner->getPets()[0]->name);
-        self::assertSame('fish', $petOwner->getPets()[0]->type);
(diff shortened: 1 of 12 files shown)


tests/AutoMapperTest.php
```diff
@@ -839,13 +839,11 @@ public function testAdderAndRemoverWithInstance(): void
...
 
         self::assertIsArray($petOwner->getPets());
-        self::assertCount(3, $petOwner->getPets());
-        self::assertSame('Nemo', $petOwner->getPets()[0]->name);
-        self::assertSame('fish', $petOwner->getPets()[0]->type);
-        self::assertSame('Félix', $petOwner->getPets()[1]->name);
-        self::assertSame('cat', $petOwner->getPets()[1]->type);
-        self::assertSame('Coco', $petOwner->getPets()[2]->name);
-        self::assertSame('dog', $petOwner->getPets()[2]->type);
+        self::assertCount(2, $petOwner->getPets());
```
joelwurtz, line 842: poke @Korbeil is this acceptable ? should we consider this PR as a BC break given that i had to change that or do we consider this a bug ?

joelwurtz: I will merge this, we can always revert before release, i need this for another PR

Korbeil: I do think it's fine since it's kinda a bug that we are adding without removing nowadays.

PR #266, opened 2025-04-08
feat(version): use composer to get installed versions instead of const

joelwurtz:
This will make things easier when releasing

### DIFF
--- a/src/Metadata/MapperMetadata.php
+++ b/src/Metadata/MapperMetadata.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace AutoMapper\Metadata;
 
-use AutoMapper\AutoMapper;
 use Composer\InstalledVersions;
 
 class MapperMetadata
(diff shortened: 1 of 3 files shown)


src/Metadata/MapperMetadata.php
```diff
@@ -67,7 +68,7 @@ public function getHash(): string
             }
         }
 
-        $hash .= AutoMapper::VERSION;
+        $hash .= InstalledVersions::getVersion('jolicode/automapper');
```
jdreesen, line 71: I think you should add `composer-runtime-api` to the `require` section of your `composer.json` when using this.

joelwurtz: Thanks didn't know that

Korbeil (approved):
Should we remove the const ? Or at least deprecate it and remove it on next major ?

joelwurtz:
Not sure a lot of people rely on this ? If we keep it, we have to update it also :/

joelwurtz:
After discussing internally, we prefer to have this to break if people rely on it, it should have been internal, but don't think many people rely on this, also since we will not update it we prefer this to break instead (so there is no silent bug)

Issue #271, opened 2025-04-30
Circular references with promoted public properties cause infinite loops


SanderVerkuil:
Assuming we have the following code:

```php

class Category
{
    /** @var Post[] */
    public array $posts = [];

    public function __construct(
        public string $name
    ) {
    }
}

class Post
{
    public function __construct(
        public string $name,
        public Category $category
    ) {
    }
}

class CategoryDTO
{
    /**
     * @var PostDTO[]
     */
    public array $posts = [];

    public function __construct(
        public string $name,
    ) {
    }
}
class PostDTO
{
    public function __construct(
        public string $name,
        public CategoryDTO $category,
    )  {
    }
}
```

the generated mapper becomes:

```php
/** @param \AutoMapper\Tests\Fixtures\Category $value */
    public function &map($value, array $context = []): mixed
    {
        if (null === $value) {
            return $value;
        }
        $sourcehash = spl_object_hash($value) . 'AutoMapper\Tests\Fixtures\CategoryDTO';
        if (\AutoMapper\MapperContext::shouldHandleCircularReference($context, $sourcehash)) {
            return \AutoMapper\MapperContext::handleCircularReference($context, $sourcehash, $value);
        }
        /** @var \AutoMapper\Tests\Fixtures\CategoryDTO|null $result */
        $result = $context['target_to_populate'] ?? null;
        if (null === $result) {
            $constructargs = [];
            if (isset($value->name)) {
                $constructargs['name'] = $value->name;
            } elseif (\AutoMapper\MapperContext::hasConstructorArgument($context, 'AutoMapper\Tests\Fixtures\CategoryDTO', 'name')) {
                $constructargs['name'] = \AutoMapper\MapperContext::getConstructorArgument($context, 'AutoMapper\Tests\Fixtures\CategoryDTO', 'name');
            } else {
                $constructargs['name'] = throw new \AutoMapper\Exception\MissingConstructorArgumentsException('Cannot create an instance of "AutoMapper\Tests\Fixtures\CategoryDTO" from mapping data because its constructor requires the following parameters to be present : "$name".', 0, null, ['name'], 'AutoMapper\Tests\Fixtures\CategoryDTO');
            }
            $result = new \AutoMapper\Tests\Fixtures\CategoryDTO(...$constructargs);
        }
        else {
            $context = \AutoMapper\MapperContext::withReference($context, $sourcehash, $result);
            $context = \AutoMapper\MapperContext::withIncrementedDepth($context);
[...]

PR #272, opened 2025-04-30
fix(constructor): fixed circular references with promoted properties

SanderVerkuil:
Updates the generated mapper code such that circular references are handled properly when using promoted properties.

Solves #271


SanderVerkuil:
@Korbeil Added the item to the changelog!

SanderVerkuil:
@Korbeil Fixed!

PR #286, opened 2025-07-17
Optimize condition order to avoid unnecessary method calls in AutoMapper

tpatartmajeur:
This PR reorders the generated condition checks in the generate() method to prioritize the customCondition() check before other standard conditions.

In the generated mapper code, this change allows custom condition to occur earlier. As a result, potentially heavy methods, such as those that may trigger database queries are not executed if the custom condition fails.

Before:
```php
if (isAllowedAttribute(...) && groupsCheck(...) && customCondition(...)) { // DB hit
   $result->someProperty = $value->someMethod(); // DB hit
}
```

After:
```php
if (customCondition(...) && isAllowedAttribute(...) && groupsCheck(...)) {
    $result->someProperty = $value->someMethod(); // DB hit only if needed
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


tpatartmajeur:
@Korbeil Done

Korbeil:
Thanks for your contribution @tpatartmajeur ! :pray:

Issue #195, opened 2024-10-08
Handle array value casting


nlemoine:
Hello, 

This is more a question than a bug report.

Are array shapes supposed to be casted? To be more accurate:

```php
<?php

class Dto
{
    public float $age;
}
$source = new Dto();
$target = $autoMapper->map(['age' => '10'], $source);

assert($target->age === 10.0); // ✅

class DtoArray
{
    /**
     * @var array<float>
     */
    public array $age;
}
$source = new DtoArray();
$target = $autoMapper->map(['age' => ['10']], $source);

assert($target->age === [10.0]); // ❌
// $target->age === ['10']

```

Would it be possible to handle array docblock types?

Korbeil:
Hey @nlemoine and thanks for reaching us about your issue !

I added your case in my local test suite and reproduced it. I do think this is not even a behavior we do "on purpose", I think this is something we have because we don't add the `declare(strict_types=1);` on top of the generated mappers.

I would say this is not something we want to support (array value casting) since we don't even support the "normal" casting and it's just a coincidence it is working at the moment.

Also, at the moment we do not add it, but since https://github.com/jolicode/automapper/pull/180 you can have it in your mappers (it's a configuration that is `false` by default). When I added it onto your example I had the following result:
```
$ ./vendor/bin/phpunit --filter=testIssue195
PHPUnit 9.6.21 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 00:00.021, Memory: 14.00 MB

There was 1 error:

1) AutoMapper\Tests\AutoMapperTest::testIssue195
TypeError: Cannot assign string to property AutoMapper\Tests\Dto::$age of type float

/home/bleduc/Sites/korbeil/automapper/tests/cache/Mapper_array_AutoMapper_Tests_Dto.php:22
/home/bleduc/Sites/korbeil/automapper/src/AutoMapper.php:116
/home/bleduc/Sites/korbeil/automapper/tests/AutoMapperTest.php:1612
```
Since this is a configuration I would like to flip from `false` to `true` by default on next major release, your behavior won't even work anymore :confused: 

@joelwurtz @nikophil what do you think about this issue ?

Baptiste

MrMeshok:
Hi everyone, I did have a thought about this issue earlier but didn't find the time to work on implementation. I think in theory we can extract array shape from doc block and create anonymous function to cast array values. It would look something like this:
```php
$cast = static fn (float $value) => $value;
$values = [];
[...]

PR #330, opened 2026-03-06
fix(transformer): cast inner values of array properties when mapping from untyped sources

nlemoine:
Hi @joelwurtz, 

I finally got the chance to test the feature addressing [my original issue](https://github.com/jolicode/automapper/issues/195). Works great using attributes, thanks again!. However, I just wondered if that could be done using phpdoc only. Because having to leverage attributes feels a bit redundant for those use cases:

```php
 // $source: ['10', '12.5']

 /** @var array<float> */
 public array $age;
```

`FromTargetMappingExtractor::transformTargetType()` mirrors target types to source for untyped sources (array, stdClass, LazyMap). For `array<int>`, the source value type becomes `int` - identical to target. `BuiltinTransformer(int,
int)` sees matching types and returns input unchanged, skipping the cast.

### Solution

In `ArrayTransformerFactory`, detect when source is untyped and the collection value types are identical mirrored scalars, then override the source value type to `mixed`. This makes the chain resolve to `BuiltinTransformer(mixed,
target_type)` which generates explicit casts.

Nullable types (`array<int|null>`) are handled via manual `NullableTransformer` wrapping since `NullableType(mixed)` is impossible in Symfony TypeInfo.

Also adds the missing `bool` entry in `BuiltinTransformer::CAST_MAPPING[mixed]`.

Let me know if this needs change.

### DIFF
--- a/src/Transformer/ArrayTransformerFactory.php
+++ b/src/Transformer/ArrayTransformerFactory.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace AutoMapper\Transformer;
 
+use AutoMapper\Lazy\LazyMap;
 use AutoMapper\Metadata\MapperMetadata;
 use AutoMapper\Metadata\SourcePropertyMetadata;
 use AutoMapper\Metadata\TargetPropertyMetadata;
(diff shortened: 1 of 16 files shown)


Korbeil:
Hey @nlemoine,
I do like your contribution but I think we should put that behind a feature flag. So we can have the current behavior (with attributes) by default and if you enable the feature you have the behavior where it takes types from phpdoc.

What do you think about this way of doing it ?

nlemoine:
Hello @Korbeil,

Sure, preserving current behavior is totally legit. I'll update this soon. Note that I also have an [upcoming PR](https://github.com/jolicode/automapper/compare/main...nlemoine:automapper:feat/array-shape) to handle array shapes. I was just waiting for this one feedback.

One thing, I would push https://github.com/jolicode/automapper/pull/330/changes/ee1272a53df307b23de4787333de5f8b5f657b2c#diff-c48da6ba30bec3efb1de38fc5cf4f40129a4b3d5d1c329f0b6bde2e148855ae9R39 into the current main since it's a "bug" from actual implementation. I can submit a separate PR if needed.

joelwurtz:
> I do like your contribution but I think we should put that behind a feature flag. 

Don't think it need a feature flag for that, we have enough of them, use case / tests seems correct for me :+1: for merging this

joelwurtz (approved):

joelwurtz:
Thanks a lot @nlemoine it should have been like this from the start (current behavior is a bug for me)

nlemoine:
Awesome, thank you @joelwurtz! I'll create [a new PR for array shapes](https://github.com/jolicode/automapper/compare/main...nlemoine:automapper:feat/array-shape) within the week.

PR #331, opened 2026-03-07
[Docs] Fix syntax error in supports method return statement

natepage:
Hi there!

I'm looking into this package as I need to map objects in a recent project, and realised this small issue in the Transformers docs so thought I would submit a quick patch,

Thanks for the amazing work done here!


Korbeil (approved):

Korbeil:
Thanks for both your fixes @natepage !

PR #336, opened 2026-03-18
feat(transformer): add ArrayShapeTransformer and support for array shape

nlemoine:
Hi @joelwurtz,

#330 added casting for array<int> but that only works for homogeneous collections (one cast in a loop). Shapes like `array{name: string, age: int, score: float}` need a different approach since `ArrayShapeType::getCollectionValueType()` returns a union of all value types and loses the per-key info.

I added an ArrayShapeTransformer that generates per-key code instead of a uniform loop. Optional keys (nickname?) get an `array_key_exists` guard. Works with nested shapes, objects, DateTime, nullable fields, and collections inside shapes. Also handles object-to-object mapping when source and target shapes have different types.

Let me know if changes are needed.

### DIFF
--- a/src/Transformer/ArrayTransformerFactory.php
+++ b/src/Transformer/ArrayTransformerFactory.php
@@ -48,7 +48,7 @@ final class ArrayTransformerFactory implements TransformerFactoryInterface, Prio
                     $fieldSourceType = $sourceShape[$key]['type'] ?? Type::mixed();
                     [$fieldSourceType] = $this->overrideSourceCollectionType($fieldSourceType, $fieldTargetType);
                 } else {
-                    $fieldSourceType = $fieldTargetType;
+                    $fieldSourceType = $sourceShape[$key]['type'] ?? $fieldTargetType;
                 }
 
                 $newSource = $source->withType($fieldSourceType);
--- a/src/Transformer/ArrayTransformerFactory.php
+++ b/src/Transformer/ArrayTransformerFactory.php
@@ -31,44 +31,6 @@ final class ArrayTransformerFactory implements TransformerFactoryInterface, Prio
             return null;
         }
 
-        // Handle array shape types — per-key transformers (skip empty shapes from object mapping)
-        if ($targetType instanceof Type\ArrayShapeType && [] !== $targetType->getShape()) {
-            $fieldTransformers = [];
-            $sourceIsUntyped = isset($mapperMetadata->source)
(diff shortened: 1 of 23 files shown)


src/Transformer/ArrayTransformerFactory.php
```diff
@@ -31,6 +31,44 @@ public function getTransformer(SourcePropertyMetadata $source, TargetPropertyMet
...
+                }
+
+                $fieldTransformers[$key] = [
+                    'transformer' => $fieldTransformer,
+                    'optional' => $field['optional'],
+                ];
+            }
+
+            return new ArrayShapeTransformer($fieldTransformers);
+        }
```
joelwurtz: I think you should add a new factory instead, this one is mainly for collection, just set the correct priority in factory to be sure that it's executed before (also in the bundle)

src/Transformer/ArrayTransformerFactory.php
```diff
@@ -31,6 +31,44 @@ public function getTransformer(SourcePropertyMetadata $source, TargetPropertyMet
...
+                    $fieldSourceType = $sourceShape[$key]['type'] ?? Type::mixed();
+                    [$fieldSourceType] = $this->overrideSourceCollectionType($fieldSourceType, $fieldTargetType);
+                } else {
+                    $fieldSourceType = $sourceShape[$key]['type'] ?? $fieldTargetType;
+                }
+
+                $newSource = $source->withType($fieldSourceType);
+                $newTarget = $target->withType($fieldTargetType);
+
+                $fieldTransformer = $this->chainTransformerFactory->getTransformer($newSource, $newTarget, $mapperMetadata);
```
joelwurtz: I'm not sure about that, you only change the type but in fact it's a whole other field / sub property, since we mainly guess on type it should be ok, but there may be case of this giving the wrong code IMO, let's keep it for now (no need to change) don't want to over complexify this

joelwurtz:
I'm :+1: on this, can you move the factory in its own file and i think we will merge this after.

src/Transformer/ArrayTransformerFactory.php
nlemoine: Done!

nlemoine:
Thanks for your feedback @joelwurtz. I moved the array shape handling into its own factory.

joelwurtz (approved):
Really nice thanks a lot

nlemoine:
Thanks @joelwurtz! 🥳

PR #355, opened 2026-09-22
chore(changelog): prepare 10.3.1

joelwurtz:
Prepares the 10.3.1 release: moves the API Platform 5 entry out of `Unreleased` and adds the compare link.

Note that no file under `src/` changed since the 10.3.0 tag — this release only carries the widened `require-dev` constraints (`api-platform/core ^5.0`, `moneyphp/money ^4.0`), the test adjustments for API Platform 5 and the castor tooling update.


joelwurtz:
Not worth a release: no file under src/ changed since 10.3.0, so the tag would ship identical library code. The API Platform 5 entry stays under Unreleased for the next real release.

